Issue #4, opened 2019-11-05
Action plugin: email send


nomonamo:
Implement a new type of action for sending email messages from within a task's step

nomonamo:
Fixed by #19

PR #19, opened 2019-11-13
Create email plugin

w3st3ry:
Related to #4

### DIFF
--- a/pkg/plugins/builtin/email/README.md
+++ b/pkg/plugins/builtin/email/README.md
@@ -1,0 +1,34 @@
+# `email` plugin
+
+This plugin send an email.
+
+## Configuration
+
+An action of type `email` requires the following kind of configuration:
+
+```yaml
+action:
+  type: email
+  configuration:
+    # mandatory, string
+    smtp_username: {{.config.smtp.username}}
+    # mandatory, string
+    smtp_password: {{.config.smtp.password}}
+    # optional, uint (default is 25)
+    smtp_port: {{.config.smtp.port}}
+    # mandatory, string
+    smtp_hostname: {{.config.smtp.hostname}}
+    # mandatory, string
+    from: foo@ovh.cloud
+    # mandatory, string
+    to: [bar@ovh.cloud, hey@ovh.cloud]
+    # mandatory, string
+    subject: Hello from µTask
+    # mandatory, string
+    body: |
+      I love baguette
+```
+
+## Note
+
+Sensitive data should be retrieved from configstore and accessed through `{{.config.[itemKey]}}` rather than hardcoded in your template.
--- a/pkg/plugins/builtin/email/README.md
+++ b/pkg/plugins/builtin/email/README.md
@@ -20,7 +20,7 @@ action:
     smtp_hostname: {{.config.smtp.hostname}}
     # mandatory, string
     from: foo@example.org
-    # mandatory, string
+    # mandatory, string collection
     to: [bar@example.org, hey@example.org]
     # mandatory, string
     subject: Hello from µTask
--- a/pkg/plugins/builtin/email/README.md
+++ b/pkg/plugins/builtin/email/README.md
@@ -19,9 +19,9 @@ action:
     # mandatory, string
     smtp_hostname: {{.config.smtp.hostname}}
     # mandatory, string
-    from: foo@ovh.cloud
+    from: foo@example.org
     # mandatory, string
-    to: [bar@ovh.cloud, hey@ovh.cloud]
+    to: [bar@example.org, hey@example.org]
     # mandatory, string
     subject: Hello from µTask
     # mandatory, string
--- a/pkg/plugins/builtin/email/email.go
+++ b/pkg/plugins/builtin/email/email.go
@@ -1,0 +1,77 @@
+package email
+
+import (
+	"bytes"
+	"fmt"
+	"net/smtp"
+	"strings"
+	"text/template"
+
+	"github.com/ovh/utask/pkg/plugins/taskplugin"
+)
+
+// the email plugin send email
+var (
+	Plugin = taskplugin.New("email", "0.1", exec,
+		taskplugin.WithConfig(validConfig, Config{}),
+	)
+)
+
+// Config is the configuration needed to send an email
+type Config struct {
+	SMTPUsername string
+	SMTPPassword string
+	SMTPPort     uint
+	SMTPHost     string
+	From         string
+	To           []string
+	Subject      string
+	Body         string
+}
+
+const emailTemplate = `From: {{.From}}<br />
+To: {{.To}}<br />
+Subject: {{.Subject}}<br />
+MIME-version: 1.0<br />
+Content-Type: text/html; charset="UTF-8"<br />
+<br />
+{{.Body}}`
+
+func validConfig(config interface{}) error {
+	return nil
+}
+
+func exec(stepName string, config interface{}, ctx interface{}) (interface{}, interface{}, error) {
+	cfg := config.(*Config)
+
+	parameters := struct {
+		From    string
+		To      string
+		Subject string
+		Body    string
+	}{
+		cfg.From,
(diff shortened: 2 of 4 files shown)


pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+		Subject string
+		Body    string
+	}{
+		cfg.From,
+		strings.Join(cfg.To, ","),
+		cfg.Subject,
+		cfg.Body,
+	}
+
+	buffer := new(bytes.Buffer)
```
Alkorin: No need for a pointer here, the zero value for Buffer is an empty buffer ready to use
https://golang.org/pkg/bytes/#Buffer

```suggestion
	var buffer bytes.Buffer
```

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+
+	auth := smtp.PlainAuth("", cfg.SMTPUsername, cfg.SMTPPassword, cfg.SMTPHostname)
+
+	port := cfg.SMTPPort
+	if port == 0 {
+		port = 25
+	}
+
+	err := smtp.SendMail(
+		fmt.Sprintf("%s:%d", cfg.SMTPHostname, int(cfg.SMTPPort)),
```
Alkorin: Cast is not needed as cfg.SMTPPort is already an integer
```suggestion
		fmt.Sprintf("%s:%d", cfg.SMTPHostname, cfg.SMTPPort),
```

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+	Plugin = taskplugin.New("email", "0.1", exec,
+		taskplugin.WithConfig(validConfig, Config{}),
+	)
+)
+
+// Config is the configuration needed to send an email
+type Config struct {
+	SMTPUsername string   `json:"smtp_username"`
+	SMTPPassword string   `json:"smtp_password"`
+	SMTPPort     uint     `json:"smtp_port,omitempty"`
```
Alkorin:
```suggestion
	SMTPPort     uint16     `json:"smtp_port,omitempty"`
```

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+	SMTPHostname string   `json:"smtp_hostname"`
+	From         string   `json:"from"`
+	To           []string `json:"to"`
+	Subject      string   `json:"subject"`
+	Body         string   `json:"body"`
+}
+
+const emailTemplate = `From: {{.From}}<br />
```
Alkorin: <br/> are unnecessary

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+
+	port := cfg.SMTPPort
+	if port == 0 {
+		port = 25
+	}
+
+	err := smtp.SendMail(
+		fmt.Sprintf("%s:%d", cfg.SMTPHostname, int(cfg.SMTPPort)),
+		auth,
+		cfg.SMTPUsername,
```
Alkorin: This parameter is the from
```suggestion
		cfg.From,
```

pkg/plugins/builtin/email/email.go
w3st3ry: right, thanks.

pkg/plugins/builtin/email/email.go
w3st3ry: Why don't use default system based builtin value?

pkg/plugins/builtin/email/email.go
w3st3ry: right, I prefer this way too.

pkg/plugins/builtin/email/email.go
w3st3ry: 👍

pkg/plugins/builtin/email/README.md
```diff
@@ -0,0 +1,34 @@
...
+    # mandatory, string
+    smtp_username: {{.config.smtp.username}}
+    # mandatory, string
+    smtp_password: {{.config.smtp.password}}
+    # optional, uint (default is 25)
+    smtp_port: {{.config.smtp.port}}
+    # mandatory, string
+    smtp_hostname: {{.config.smtp.hostname}}
+    # mandatory, string
+    from: foo@ovh.cloud
```
Alkorin: ovh.cloud may belong to someone

RFC2606 specify domains suitable for documentation:
- example.com
- example.net
- example.org

https://tools.ietf.org/html/rfc2606#section-3

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+	SMTPHostname string   `json:"smtp_hostname"`
+	From         string   `json:"from"`
+	To           []string `json:"to"`
+	Subject      string   `json:"subject"`
+	Body         string   `json:"body"`
+}
+
+const emailTemplate = `From: {{.From}}<br />
+To: {{.To}}<br />
+Subject: {{.Subject}}<br />
```
Alkorin: Beware, email headers should not contain non-ascii characters, so subject can't contains UTF-8 characters and should be encoded as described in RFC 1342

https://tools.ietf.org/html/rfc1342

pkg/plugins/builtin/email/email.go
Alkorin: Because an TCP destination port is officially a 16 bit unsigned integer

https://tools.ietf.org/html/rfc793#page-15

Alkorin: And then, the unmarshal will fail if someone provides 99999 as TCP port instead of failing on smtp.SendMail

pkg/plugins/builtin/email/email.go
w3st3ry: Thanks for the information, fixed.

pkg/plugins/builtin/email/README.md
```diff
@@ -0,0 +1,34 @@
...
+    smtp_username: {{.config.smtp.username}}
+    # mandatory, string
+    smtp_password: {{.config.smtp.password}}
+    # optional, uint (default is 25)
+    smtp_port: {{.config.smtp.port}}
+    # mandatory, string
+    smtp_hostname: {{.config.smtp.hostname}}
+    # mandatory, string
+    from: foo@example.org
+    # mandatory, string
```
nomonamo: s/string/collection/

w3st3ry: fixed.

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,108 @@
...
+		fmt.Sprintf("%s:%d", cfg.SMTPHostname, cfg.SMTPPort),
+		auth,
+		cfg.From,
+		cfg.To,
+		buffer.Bytes())
+	if err != nil {
+		return nil, nil, fmt.Errorf("Send email failed: %s", err.Error())
+	}
+
+	return nil, nil, nil
```
nomonamo: pondering about this for a while: I think I'd return the content of the "parameters" struct as the step's output, in case any steps down the line want to re-use that info

w3st3ry: seems legit. done.

pkg/plugins/builtin/email/email.go
```diff
@@ -0,0 +1,108 @@
...
+		return errors.New("body is missing")
+	}
+
+	return nil
+}
+
+func exec(stepName string, config interface{}, ctx interface{}) (interface{}, interface{}, error) {
+	cfg := config.(*Config)
+
+	parameters := struct {
```
nomonamo: nitpicking for overall consistency: add json tags to lowercase "from", "to", etc... otherwise the output object will be keyed "From", "To", ...

w3st3ry: done.

PR #20, opened 2019-11-15
typos + broken links in documentation

nomonamo:
fix a couple of typos, update to latest version


nomonamo:
changed version in Makefile
would like to make this value a parameter to be received when generating the install script and when building with make

w3st3ry:
> changed version in Makefile
> would like to make this value a parameter to be received when generating the install script and when building with make

What about directly get the value from the latest tag?
Something like `git describe --tags $(git rev-list --tags --max-count=1)`

Makefile
```diff
@@ -6,7 +6,7 @@ TEST_LOCATION	= ./...
 TEST_CMD		= go test -v -mod=vendor -cover ${TEST_LOCATION}
 TEST_CMD_COV	= ${TEST_CMD} -covermode=count -coverprofile=coverage.out
 
-VERSION			= v1.0.0
+VERSION			= v1.0.1
```
w3st3ry: As I said about get the latest version, WDYT?
```
VERSION = `git describe --tags $(git rev-list --tags --max-count=1)`
```

nomonamo: that's looks good
however... I'm having 2nd feelings, this seems weird: if the latest tag is 5 commits ago, are you really running that version of µTask?

w3st3ry: It's the problem of the end user I think.
It always recommended to fetch the latest release instead of clone the master branch.
And if it's a tech dev want to fork or anything on his side, he can also do a `git log` to see the history upstream.

nomonamo:
LICENSE file in fizz is a mistake :/

Makefile
nomonamo: fair enough, I'll apply your proposal

hack/generate-install-script.sh
```diff
@@ -35,4 +35,8 @@ write_block "sql/schema.sql"       sql/schema.sql
 echo "### TEMPLATES" >> $dst
 write_block "templates/hello-world-now.yaml" ./examples/templates/hello-world-now.yaml
 
+version=`git describe --tags $(git rev-list --tags --max-count=1)`
+sed -i.bak "s/DOCKER_TAG/$version/" ./$dst
```
Alkorin: No need to create a backup file if it is deleted just after ;)
```suggestion
sed -i "s/DOCKER_TAG/$version/" ./$dst
```

hack/generate-install-script.sh
```diff
@@ -35,4 +35,8 @@ write_block "sql/schema.sql"       sql/schema.sql
...
+rm $dst.bak
```
Alkorin:
```suggestion
```

hack/generate-install-script.sh
nomonamo: incredible as it may sound, it's there for compatibility with mac os :|

Makefile
w3st3ry: 👯‍♂

Makefile
```diff
@@ -6,7 +6,7 @@ TEST_LOCATION	= ./...
 TEST_CMD		= go test -v -mod=vendor -cover ${TEST_LOCATION}
 TEST_CMD_COV	= ${TEST_CMD} -covermode=count -coverprofile=coverage.out
 
-VERSION			= v1.0.0
+VERSION         = `git describe --tags $(git rev-list --tags --max-count=1)`
```
w3st3ry: Equal sign should be aligned on the same column as the others.

nomonamo: fixed

hack/generate-install-script.sh
nomonamo: I've added a comment in the code, for clarity

Alkorin: The ```-i``` option is not POSIX compliant. So you can't expect it to work everywhere.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

See #22

PR #25, opened 2019-11-18
Rework email plugin

w3st3ry:
I just wrote some code to replace the "homemade" way to send e-mail.
The PR contains the updated vendors.

Also related to #4


pkg/plugins/builtin/email/email.go
```diff
@@ -1,44 +1,37 @@
...
-	SMTPPassword string   `json:"smtp_password"`
-	SMTPPort     uint16   `json:"smtp_port,omitempty"`
-	SMTPHostname string   `json:"smtp_hostname"`
-	From         string   `json:"from"`
-	To           []string `json:"to"`
-	Subject      string   `json:"subject"`
-	Body         string   `json:"body"`
+	SMTPUsername      string   `json:"smtp_username"`
+	SMTPPassword      string   `json:"smtp_password"`
+	SMTPPort          uint16   `json:"smtp_port"`
```
loopfz: Careful, we only template strings

w3st3ry: Thanks, fixed!

w3st3ry: thanks, fixed!

pkg/plugins/builtin/email/email.go
```diff
@@ -70,38 +80,43 @@ func validConfig(config interface{}) error {
...
+	message.SetHeader("To", recipients...)
+	message.SetHeader("Subject", cfg.Subject)
+	message.SetBody(
+		http.DetectContentType([]byte(cfg.Body)),
+		cfg.Body,
+	)
 
-	auth := smtp.PlainAuth("", cfg.SMTPUsername, cfg.SMTPPassword, cfg.SMTPHostname)
+	port, _ := strconv.ParseUint(cfg.SMTPPort, 10, 64)
+	skipTLS, _ := strconv.ParseBool(cfg.SMTPSkipTLSVerify)
```
nomonamo: ambiguous that this defaults to false when parsing "", WDYT? correct, but made me stop reading

pkg/plugins/builtin/email/email.go
```diff
@@ -70,38 +80,43 @@ func validConfig(config interface{}) error {
...
+	if err := d.DialAndSend(message); err != nil {
+		fmt.Errorf("Send email failed: %s", err.Error())
 	}
 
-	err := smtp.SendMail(
-		fmt.Sprintf("%s:%d", cfg.SMTPHostname, cfg.SMTPPort),
-		auth,
-		cfg.From,
+	// to reuse configuration
+	parameters := struct {
```
nomonamo: nitpicking, this struct seems redundant with part of Config, 
maybe declare it at the top and embed it in Config?

```suggestion
type mailParams struct {
	FromAddress       string   `json:"from_address"`
	FromName          string   `json:"from_name,omitempty"`
	To                []string `json:"to"`
	Subject           string   `json:"subject"`
	Body              string   `json:"body"`
}

type Config struct {
    ...
    mailParams
}
```

pkg/plugins/builtin/email/email.go
Alkorin: Do we *really* need a skipTLS parameter in 2019-2020 ?

pkg/plugins/builtin/email/email.go
```diff
@@ -70,38 +80,43 @@ func validConfig(config interface{}) error {
...
-		strings.Join(cfg.To, ","),
-		mime.BEncoding.Encode("UTF-8", cfg.Subject),
-		cfg.Body,
-	}
+	message := mail.NewMessage()
 
-	buffer := new(bytes.Buffer)
+	recipients := make([]string, len(cfg.To))
+	for i, recipient := range cfg.To {
+		recipients[i] = message.FormatAddress(recipient, "")
```
Alkorin: What is the goal of calling ```message.FormatAddress``` ?

https://github.com/go-mail/mail/blob/v2.3.1/message.go#L135

if name is empty, it returns the address :)

pkg/plugins/builtin/email/email.go
```diff
@@ -70,38 +80,43 @@ func validConfig(config interface{}) error {
...
 
-	auth := smtp.PlainAuth("", cfg.SMTPUsername, cfg.SMTPPassword, cfg.SMTPHostname)
+	port, _ := strconv.ParseUint(cfg.SMTPPort, 10, 64)
+	skipTLS, _ := strconv.ParseBool(cfg.SMTPSkipTLSVerify)
 
-	port := cfg.SMTPPort
-	if port == 0 {
-		port = 25
+	d := mail.NewDialer(cfg.SMTPHostname, int(port), cfg.SMTPUsername, cfg.SMTPPassword)
+	d.TLSConfig = &tls.Config{InsecureSkipVerify: skipTLS}
```
Alkorin: You override the default config which set the ServerName to the HostName

https://github.com/go-mail/mail/blob/v2.3.1/smtp.go#L149

```suggestion
	d.TLSConfig = &tls.Config{InsecureSkipVerify: skipTLS, ServerName: cfg.SMTPHostname}
```

But do we really need an InsecureSkipVerify option ... ;)

pkg/plugins/builtin/email/email.go
```diff
@@ -70,38 +80,43 @@ func validConfig(config interface{}) error {
...
+	if err := d.DialAndSend(message); err != nil {
+		fmt.Errorf("Send email failed: %s", err.Error())
```
Alkorin: The error is "silently" dropped and printed to stderr which is maybe not collected by the application owner.

Can we provide a custom logger to propper collect the error if it is not returned?

README.md
```diff
@@ -422,7 +422,14 @@ Browse [builtin actions](./pkg/plugins/builtin)
 |              || `smtp_password`: password of SMTP server
 |              || `smtp_port`: port of SMTP server
 |              || `smtp_hostname`: hostname of SMTP server
-|              || `from`: from which email you want to send the message
+|**`email`**   | Send an email
```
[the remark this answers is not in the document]
w3st3ry: yes, binding issue...

pkg/plugins/builtin/email/email.go
w3st3ry: Agree! fixed.

pkg/plugins/builtin/email/email.go
w3st3ry: thanks!

pkg/plugins/builtin/email/email.go
w3st3ry: @nomonamo I added an explicit comment.
@Alkorin I think it's an usecase for testing purpose.

pkg/plugins/builtin/email/email.go
w3st3ry: fixed, thanks!

pkg/plugins/builtin/email/email.go
```diff
@@ -82,42 +86,32 @@ func exec(stepName string, config interface{}, ctx interface{}) (interface{}, in
...
+	params := &mailParameters{
 		cfg.FromAddress,
 		cfg.FromName,
 		cfg.To,
 		cfg.Subject,
 		cfg.Body,
 	}
 
-	return &parameters, nil, nil
+	return params, nil, nil
```
nomonamo: I believe you can just return cfg.mailParameters :)

pkg/plugins/builtin/email/email.go
nomonamo: I know it might seem obvious, but I'd mention what the default value is in case of err != nil, even if that's "not possible" per preemptive validation

```suggestion
port, _ := strconv.ParseUint(cfg.SMTPPort, 10, 64) // defaults to 0
skipTLS, _ := strconv.ParseBool(cfg.SMTPSkipTLSVerify) // defaults to false
```

which begs the question, false seems ok, but is 0 an appropriate default?

pkg/plugins/builtin/email/email.go
w3st3ry: my bad, thanks.

pkg/plugins/builtin/email/email.go
w3st3ry: I added default comments.
BTW port can't be not set default because https://golang.org/src/strconv/atoi.go?s=1789:1852#L61 generate an error in `validConfig` func (https://github.com/ovh/utask/pull/25/files#diff-7e310a8e891eb8f9cb73c9279c5a894bR51).

PR #37, opened 2019-11-27
Improve code coverage

w3st3ry:
Signed-off-by: Valentin Pichard <7628998+w3st3ry@users.noreply.github.com>


ovh-cds:
CDS Report test#145.0 ✘  
* Stage 1
  * Test ✘

api/api_test.go
```diff
@@ -64,12 +64,24 @@ func TestMain(m *testing.M) {
...
 	hdl = srv.Handler(ctx)
 
 	os.Exit(m.Run())
 }
 
 func dumbIdentityProvider(r *http.Request) (string, error) {
+	username := r.Header.Get(usernameHeaderKey)
+
+	if username != adminUser && username != regularUser {
+		return "", errors.New("unknow user")
```
rbeuque74:
```suggestion
		return "", errors.New("unknown user")
```

Typo here

ovh-cds:
CDS Report test#147.0 ✘  
* Stage 1
  * Test ✘

api/api_test.go
w3st3ry: fixed

api/api_test.go
```diff
@@ -99,9 +111,44 @@ func TestUtils(t *testing.T) {
...
+			iffy.ExpectStatus(301),
+		)
+
+	tester.AddCall("testMeta", http.MethodGet, "/meta", "").
+		Headers(regularHeaders).
+		Checkers(
+			iffy.ExpectStatus(200),
+		)
+
+	utask.FMaintenanceMode = true
```
nomonamo: be careful, this is misleading: the calls are declared with `tester.AddCall`, but they are only performed on `tester.Run()`
this means that `FMaintenanceMode` is effectively `true` for every call in this test
I would place this declaration either at the beginning of the test, with a deferred set to false
or right before `tester.Run()`, whatever seems more readable to you

w3st3ry: Fixed!
Thanks

Issue #38, opened 2019-11-28
Action plugin: script


loopfz:
Implement a plugin "script" that forks and executes a script within the utask container.

**Suggestion:**

```action:
    type: script
    configuration:
        file: check-invoice.py
        argv:
            - {{.step.getInvoice.output.invoiceID}}
            - ...
        stdin: foobar
        timeout: 5s
```

This should execute for the step the script **check-invoice.py**, which needs to be present in a configured _script folder_ within the utask container.
The script should be executable (+x), and use any interpreter as long as it is present in the utask container. The base utask image should maybe provide the most common interpreters?

There should be a way to pass data to the following steps in a structured way, it may adopt a similar strategy as the ssh plugin: the _last_ line of the script's stdout will be json parsed, and if it succeeds will be used as the result payload for the step.

w3st3ry:
Merged #46

PR #46, opened 2019-12-03
feat: Implement script plugin

w3st3ry:
Related to #38


pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+	lastNL := strings.LastIndexByte(outStr, '{')
+	if lastNL == -1 {
+		return nil, metadata, nil
+	}
+
+	lastLine := out[lastNL:]
+	payload := make(map[string]interface{})
+	err = json.Unmarshal([]byte(lastLine), &payload)
+	if err != nil {
+		return nil, nil, err
```
loopfz: what if the script doesnt return actual json?
should we be more permissive and silently skip the error? or perhaps make this configurable (ie, no_output = true to avoid the unmarshal attempt)

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+	cfg := config.(*Config)
+
+	var timeout time.Duration
+
+	if cfg.Timeout != "" {
+		t, _ := strconv.ParseInt(cfg.Timeout, 10, 64)
+		timeout = time.Duration(t)
+	} else {
+		// default is 5*60 = 300 seconds or 5 minutes
+		timeout = time.Duration(300)
```
loopfz: default should probably be lower than 5 minutes

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func exec(stepName string, config interface{}, ctx interface{}) (interface{}, interface{}, error) {
+	cfg := config.(*Config)
+
+	var timeout time.Duration
+
+	if cfg.Timeout != "" {
+		t, _ := strconv.ParseInt(cfg.Timeout, 10, 64)
+		timeout = time.Duration(t)
```
loopfz: should we parse a duration string instead?
ie 5s, 1m, ...

loopfz: additionally, it's called "timeout_seconds" but is actually in ms

pkg/plugins/builtin/script/script.go
loopfz: but this is actually 300ms

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+
+func validConfig(config interface{}) error {
+	cfg := config.(*Config)
+
+	if cfg.File == "" {
+		return errors.New("file is missing")
+	}
+
+	if cfg.Timeout != "" {
+		if _, err := strconv.ParseUint(cfg.Timeout, 10, 64); err != nil {
```
loopfz: keep in mind that this doesnt allow templating for this field, since {{.config.whatever}} wont pass ParseUint.
Probably acceptable in this specific case.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+		exitCode = cmd.ProcessState.Sys().(syscall.WaitStatus).ExitStatus()
+	}
+
+	signal := cmd.ProcessState.Sys().(syscall.WaitStatus).Signal().String()
+
+	outStr := string(out)
+
+	metadata := struct {
+		ExitCode   string `json:"exit_code"`
+		ExitSignal string `json:"exit_signal"`
```
loopfz: can I have the exec duration ?

loopfz: would probably be useful in other plugins, too

loopfz: maybe we could even insert it from within the engine, with a reserved name

cmd/utask/root.go
```diff
@@ -129,6 +136,19 @@ var rootCmd = &cobra.Command{
...
 		}
 
+		sFiles, err := ioutil.ReadDir(utask.FScriptsFolder)
+		if err != nil {
+			log.Warnf("Ignoring scripts directory %s: %s", utask.FScriptsFolder, err)
+		}
+
+		if len(sFiles) > 0 {
+			for _, f := range sFiles {
+				if f.Mode()&0010 == 0 && f.Name()[0] != '.' {
```
loopfz: should we check the type? ie, regular file or symlink, to avoid folders, named pipes, etc?

loopfz: also, 010 seems weird

ovh-cds:
CDS Report test#176.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#177.0 ✘  
* Stage 1
  * Test ✘

pkg/plugins/builtin/script/script.go
w3st3ry: I switch to 2minutes. And it's not 300ms, you missed the conversion. I moved the conversion for more readability.

pkg/plugins/builtin/script/script.go
w3st3ry: I think it's more readable like this. WDYT @nomonamo ?
It's not in ms, you missed the conversion. I moved the conversion for more readability.

ovh-cds:
CDS Report test#178.0 ✘  
* Stage 1
  * Test ✘

cmd/utask/root.go
w3st3ry: We don't do that. I think it's the responsibility of the instance admin.
`010` match with rights of users belonging to the group. You think owners is needed too?

pkg/plugins/builtin/script/script.go
w3st3ry: I switched with `process_state`, I think it's more relevant. I will checkout about exec duration.

pkg/plugins/builtin/script/script.go
w3st3ry: Why? from the moment we receive a correct string who contains a number > 0 it will works.

pkg/plugins/builtin/script/script.go
w3st3ry: I fixed that by returns metada in any case. I will check about add `no_output` option.

ovh-cds:
CDS Report test#179.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#180.0 ✘  
* Stage 1
  * Test ✘

engine/templates_tests/execScript.yaml
```diff
@@ -0,0 +1,18 @@
...
+- name: file
+  description: file path
+
+steps:
+    stepOne:
+        description: first step
+        action:
+            type: script
+            configuration:
+                file: "{{.input.file}}"
```
fsamin: we shouldn't allow this.
A template should be quite determinist, and this use case sounds like a "meta-template" and is a possible vulnerability.

pkg/plugins/builtin/script/script.go
loopfz: validConfig is called without any templated values substitution

pkg/plugins/builtin/script/script.go
loopfz: Ok, I see the conversion down below. This is a lot _less_ readable in my opinion, since all the code paths manipulating a time.Duration before that seem to be buggy (ie, the comment I made).

pkg/plugins/builtin/script/script.go
loopfz: yeah, if the script is not supposed to return json that unmarshal is wasteful at best, dangerous at worst

pkg/plugins/builtin/script/script.go
loopfz: I'd even go down to 30s

cmd/utask/root.go
loopfz: @fsamin you make a good point, but still if something is wrong with my utask packaging, I'd rather know right away (and maybe have my orchestrator not go through with the update because it panics at boot), than find out a part of my workflows are not working later.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,108 @@
...
+		timeout = time.Duration(t)
+	} else {
+		// default is 5*60 = 300 seconds or 5 minutes
+		timeout = time.Duration(300)
+	}
+
+	ctxe, cancel := context.WithTimeout(context.Background(), timeout*time.Second)
+	defer cancel()
+
+	cmd := gexec.CommandContext(ctxe, cfg.File, cfg.Argv...)
```
loopfz: Shouldn't cfg.File be prefixed with the scripts folder ?

README.md
```diff
@@ -391,7 +391,7 @@ Browse [builtin actions](./pkg/plugins/builtin)
 |**`ssh`**     | Connect to a remote system and run commands on it | [Access plugin doc](./pkg/plugins/builtin/ssh/README.md)
 |**`email`**   | Send an email | [Access plugin doc](./pkg/plugins/builtin/email/README.md)
 |**`ping`**    | Send a ping to an hostname *Warn: This plugin will keep running until the count is done* | [Access plugin doc](./pkg/plugins/builtin/ping/README.md)
-
+|**`script`**    | Execute a script under `scripts` folder *Warn: This plugin will keep running until execution is done* | [Access plugin doc](./pkg/plugins/builtin/script/README.md)
```
loopfz: I don't understand the warning. There is a timeout option, and this is not any more true than with other plugins e.g. HTTP ?

cmd/utask/root.go
fsamin: IMO utask should check script files when it is parsing the templates... but not check ALL the files in this directory
BTW if the script is invalid (eg. syntax error) the workflows won't work, then we have to handle this error a nice way within the web ui

loopfz: @fsamin maybe the script plugin's **ValidConfig()** could check file existence and +x ? That way only the used script files would be checked. And afaik **ValidConfig()** would be called in that code path, because it is parsed + validated before being updated? cc @nomonamo confirm ?

loopfz: Validating script files that way (via **ValidConfig()**) would prevent the use of templating for the _file_ field, as a byproduct (because _{{.input.foobar}}_ is probably not an existing file).
This might not be a drawback, re: the discussion started below by @fsamin

pkg/plugins/builtin/script/script.go
w3st3ry: I don't know why, but you comment outdated code.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func validConfig(config interface{}) error {
+	cfg := config.(*Config)
+
+	if cfg.File == "" {
+		return errors.New("file is missing")
+	}
+
+	if cfg.Timeout != "" {
```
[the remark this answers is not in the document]
w3st3ry: Yes we did that but not in `validConfig()`. Checkout `exec()` func.

engine/templates_tests/execScript.yaml
w3st3ry: This is only a test template used for internal tests, not an example.

README.md
w3st3ry: This warn is the same as `ping` plugin because it's not just about ms, it can be several seconds.
But maybe you right, I should remove it.

engine/templates_tests/execScript.yaml
fsamin: can you confirm that we don't support this feature ?

loopfz: right now it works (naturally), even though it's dumb.
see above discussion around **ValidConfig()** suggestion.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,114 @@
...
+	cfg := config.(*Config)
+
+	var timeout time.Duration
+
+	if cfg.Timeout != "" {
+		t, _ := strconv.ParseInt(cfg.Timeout, 10, 64)
+		timeout = time.Duration(t)
+	} else {
+		// default is 2*60 = 120 seconds or 2 minutes
+		timeout = time.Duration(120)
```
nomonamo: IMO this is more idiomatic:
```suggestion
		timeout = 120 * time.Second
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,114 @@
...
+	exitCode := 0
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		if exitError, ok := err.(*gexec.ExitError); ok {
+			exitCode = exitError.Sys().(syscall.WaitStatus).ExitStatus()
+		} else {
+			exitCode = 1
+			if string(out) == "" {
+				out = []byte(err.Error())
```
nomonamo: what happens to `err` if `string(out) != ""` ?

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,114 @@
...
+
+	outStr := string(out)
+
+	metadata := Metadata{
+		ExitCode:     fmt.Sprint(exitCode),
+		ProcessState: pState,
+		Output:       outStr,
+	}
+
+	lastNL := strings.LastIndexByte(outStr, '{')
```
nomonamo: I'm unsure how this is supposed to turn out, what if outStr consists of:
```
{"id":1234,"properties":{"foo":"bar"}}
```
would you start parsing right after  `"properties"` ?

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,114 @@
...
+		// default is 2*60 = 120 seconds or 2 minutes
+		timeout = time.Duration(120)
+	}
+
+	timeout = timeout * time.Second
+
+	ctxe, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	cmd := gexec.CommandContext(ctxe, utask.FScriptsFolder+cfg.File, cfg.Argv...)
```
nomonamo: don't we risk missing a `/` between folder and file?
intuitively, I'd pass the folder like this in my env:
```
SCRIPTS=./my_scripts_folder
```

ovh-cds:
CDS Report test#184.0 ✘  
* Stage 1
  * Test ✘

pkg/plugins/builtin/script/script.go
loopfz: doesnt work indeed. this code was lifted from the ssh package, which had a naive implementation (and only dealt with scalars)

ovh-cds:
CDS Report test#185.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#186.0 ✘  
* Stage 1
  * Test ✘

engine/engine_test.go
```diff
@@ -555,6 +557,29 @@ func TestBaseOutput(t *testing.T) {
 	assert.Equal(t, "bar", output["foo"])
 }
 
+func TestScriptPlugin(t *testing.T) {
+	file := os.Getenv("SCRIPTS") + "/hello-world.py"
```
loopfz: The test should probably use /bin/sh instead of python to be runnable everywhere

Alkorin:
```suggestion
	file := filepath.Join(os.Getenv("SCRIPTS"), "hello-world.py")
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,125 @@
...
+		timeout, _ = time.ParseDuration(cfg.Timeout)
+	} else {
+		// default is 2 * 1 minute = 2 minutes
+		timeout = time.Duration(2) * time.Minute
+	}
+
+	ctxe, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	cmd := gexec.CommandContext(ctxe, utask.FScriptsFolder+cfg.File, cfg.Argv...)
```
Alkorin:
```suggestion
	cmd := gexec.CommandContext(ctxe, filepath.Join(utask.FScriptsFolder, cfg.File), cfg.Argv...)
```

pkg/plugins/builtin/script/script.go
w3st3ry: I added exec time in the plugin itself for now. I'll do another PR in the future to integrate this in the core engine directly.

cmd/utask/root.go
```diff
@@ -31,12 +31,14 @@ const (
 	defaultInitializersFolder = "./init"
 	defaultPluginFolder       = "./plugins"
 	defaultTemplatesFolder    = "./templates"
+	defaultScriptsFolder      = "./scripts/"
```
nomonamo: uneasy about this trailing slash, IMO should be handled in code

engine/engine_test.go
```diff
@@ -555,6 +557,35 @@ func TestBaseOutput(t *testing.T) {
...
+	res, err := createResolution("execScript.yaml", nil, nil)
+	assert.NotNil(t, res)
+	assert.Nil(t, err)
+
+	res, err = runResolution(res)
+	assert.NotNil(t, res)
+	assert.Nil(t, err)
+
+	payload := make(map[string]interface{})
+	payload["msg"] = "Hello world!"
```
nomonamo: one silly suggestion: make `world` an input to the template, 
instead of making the argv a fixed feature in the template execScript.yaml
defining that value in the test would make it more readable and robust, imho

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,142 @@
...
+		timeout, _ = time.ParseDuration(cfg.Timeout)
+	} else {
+		// default is 2 * 1 minute = 2 minutes
+		timeout = time.Duration(2) * time.Minute
+	}
+
+	ctxe, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	cmd := gexec.CommandContext(ctxe, filepath.Join(utask.FScriptsFolder, cfg.File), cfg.Argv...)
```
fsamin: What about the working directory of the command ?
If you set if, it will inherit from the working directory of the parent process.
IMO we should set it to the directory of `filepath.Join(utask.FScriptsFolder, cfg.File)`. It makes the behavior more predictable for a script developer.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,142 @@
...
+	ExitCode      string `json:"exit_code"`
+	ProcessState  string `json:"process_state"`
+	Output        string `json:"output"`
+	ExecutionTime string `json:"execution_time"`
+	Error         string `json:"error"`
+}
+
+// Config is the configuration needed to execute a script
+type Config struct {
+	File    string   `json:"file,required"`
```
fsamin: Another proposition:

not sure about `file`. It could be `command`. And i would be able to set it this way:
```
command: python myScript.py arg1 arg2 arg3
```
or
```
command: bash -C "cd myScript && pip install && python main.py arg1 arg2 arg3"
```
To run it we have to split this string, consider slice[0] as the command and slice[1:] its arguments .
This way we juste have to check that the command valid: https://golang.org/pkg/os/exec/#LookPath

Alkorin: Why split ? Never try to split on " " to execute a command, it will fail.

It should be written that way:
command: [ "bash", "-c", "cd myScript && pip install && python main.py arg1 arg2 arg3"]

loopfz: @fsamin I think your example perfectly shows why it should be _file_ and not an arbitrary shell command: we want to treat these scripts as black boxes with as little logic as possible in the step configuration.
The setup should be done elsewhere (which interpreter = shebang, pip install = definitely not at each step exec, but one-shot at Dockerfile?).
The current directory is interesting and may warrant a separate, generic config field.

engine/templates_tests/execScript.yaml
```diff
@@ -0,0 +1,15 @@
...
+title_format: "[test] a simple task"
+
+steps:
+    stepOne:
+        description: first step
+        action:
+            type: script
+            configuration:
+                # Only for testing purpose
+                file: ./../scripts/hello-world.sh
```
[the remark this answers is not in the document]
loopfz: agree. I'd even argue that using **..** in the file param should be prevented ?

cmd/utask/root.go
loopfz: agree

pkg/plugins/builtin/script/script.go
w3st3ry: Fixed.

pkg/plugins/builtin/script/script.go
w3st3ry: Checked /w @fsamin.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,143 @@
...
+	Argv    []string `json:"argv,omitempty"`
```
loopfz: I think stdin should be configurable

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,143 @@
...
+	lastLine := ""
+
+	for i := len(outputArray) - 1; i >= 0; i-- {
+		if len(outputArray[i]) > 0 {
+			lastLine = outputArray[i]
+			break
+		}
+	}
+
+	if lastLine == "" {
```
loopfz: if it's empty, OR no json found ?
I really dislike trying to unmarshal something that's predictably not json.
Ideally, configurable. But detecting the presence of {} may be worthwhile too.

scripts/hello-world.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+echo 'Hello world script'
+echo '{"msg":"Hello '$1'!","error":{"message":"nil"}}'
```
loopfz: this will probably serve as a goto example of how to output json from a .sh step
should we use

cat <<EOF
{"msg": "Hello $1!"}
EOF

? Less escaping tricks.

Also the error object here is a bit confusing, it resembles some form of contract for error handling but I see no other reference to it in the plugin code or doc.

Alkorin: As a best practice,you should never try to create structured data by string concatenation, it doesn't work well and can conduct to injection issues.

You should use tool who knows how to handle JSON like jq
```
$ echo '"John" Doe' | jq -R '{"msg":("Hello " + . + "!"), "error":{"message":null}}'
{
  "msg": "Hello \"John\" Doe!",
  "error": {
    "message": null
  }
}
```

engine/templates_tests/execScript.yaml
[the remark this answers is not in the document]
w3st3ry: Checked IRL /w @nomonamo.
I moved `hello-world.sh` script outside from `scripts` folder because it's not an example but just a testing addon.

scripts/hello-world.sh
w3st3ry: I agree /w both of you but I moved this script in a dedicated testing folder to highligth the fact that is not an example, just a dumb script wrote *for testing purpose*. So did really we need changes on it? WDYT?

@Alkorin I also agree about using `jq` for this kind of stuff but I don't want to embed a system dependency in each testing env (CI/laptop/desktop...), it works fine in this way *for testing purpose*.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,147 @@
...
+	if cfg.File == "" {
+		return errors.New("file is missing")
+	}
+
+	f, err := os.Stat(filepath.Join(utask.FScriptsFolder, cfg.File))
+	if os.IsNotExist(err) {
+		return fmt.Errorf("%s not found in FS: %s", cfg.File, err.Error())
+	}
+
+	if f.Mode()&0010 == 0 {
```
Alkorin: Why checking `g+x` and not `o+x` or `u+x` ?
 
```suggestion
	if f.Mode()&0111 == 0 {
```

Will tell you if it's executable by somebody

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,147 @@
...
+	lastLine := ""
+
+	for i := len(outputArray) - 1; i >= 0; i-- {
+		if len(outputArray[i]) > 0 {
+			lastLine = outputArray[i]
+			break
+		}
+	}
+
+	if lastLine == "" &&
```
Alkorin: Something wrong here, if `lastLine` is empty, `!strings.Contains(lastLine, "{")` and `!strings.Contains(lastLine, "}")` will always be true.

strings.Contains already have a shortcut for len == 0

Should be something like:
```
	if lastLine == "" || !(strings.Contains(lastLine, "{") && strings.Contains(lastLine, "}") {
		return nil, metadata, nil
	}
```

If `lastLine` is empty OR lastLine doesn't contains `{` AND `}`.

But strings.Contains on an empty string is almost free, so `lastLine == ""` looks like some premature optimization and for clarity it is maybe better to write it like this:

```
	// If it doesn't look like JSON
	if !(strings.Contains(lastLine, "{") && strings.Contains(lastLine, "}") {
		return nil, metadata, nil
	}
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,147 @@
...
+	cmd.Dir = utask.FScriptsFolder
+	cmd.Stdin = strings.NewReader(cfg.Stdin)
+
+	exitCode := 0
+	metaError := ""
+
+	// start exec time timer
+	timer := time.Now()
+	// execute script
+	out, err := cmd.CombinedOutput()
```
Alkorin: Why do you want to combine Stdout & Stderr?

As you may expect some JSON as script output, it is maybe better to keep them separated because there is no guarantee on how they are combined

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if lastLine == "" &&
+		!strings.Contains(lastLine, "{") &&
+		!strings.Contains(lastLine, "}") {
+		return nil, metadata, nil
+	}
+
+	payload := make(map[string]interface{})
+	err = json.Unmarshal([]byte(lastLine), &payload)
```
Alkorin: Shouldn't it be some config to enable parsing of stdout instead of trying to guess it?

It would allow scripts to write pretty-printed json

pkg/plugins/builtin/script/script.go
w3st3ry: Because we only care if we're able to execute it based on group owner.
There is some cases you're not the main owner but the user belong to the group owner.
We also don't care about other users permissions.

pkg/plugins/builtin/script/script.go
w3st3ry: Right, thanks for this typo!

pkg/plugins/builtin/script/script.go
w3st3ry: Actually not, we prefer to use the same kind of 'interface' as SSH plugin to retrieve something standard.
The deal is clear and simple as maintainer and as scripts writers, so for the moment we want to keep it as that.

pkg/plugins/builtin/script/script.go
w3st3ry: It's the responsibility of the script's writer to handle this kind of issue.
Otherwise, if you want to debug, as discussed /w @loopfz, you want to keep a consistent trace of the program execution.

pkg/plugins/builtin/script/script.go
Alkorin: In fact, all you can check is that the script is at least executable, by someone.

Here you check that the group associated to the file can execute it, but you don't check that the user belong to that group. So the check is useless.

Here your test will fail but I can exec the script:
```
toms@thor:~/utask$ ls -alhtr a.sh
-rwxr--r-- 1 toms toms 0 Dec 15 19:36 a.sh
toms@thor:~/utask$ ./a.sh
toms@thor:~/utask$
```

Here your test will succeed but I can't exec the script:
```
toms@thor:~/utask$ ls -alhtr b.sh
-rw-r-xr-- 1 toms root 0 Dec 15 19:36 b.sh
toms@thor:~/utask$ ./b.sh
-bash: ./b.sh: Permission denied
toms@thor:~/utask$
```
Either check that the file is at least executable by someone, or do the full permission check

pkg/plugins/builtin/script/script.go
Alkorin: It can't be the responsability of the script's writer as he can't control how YOU merge the two streams and then could conduct to unconsistant behaviour.

If the script output one line to stdout and one line to stderr, you will either have stdout first and stderr last or the opposite.

So it is better to split the two streams (and fix the ssh plugin if it does the same) or you should document that the script should write either to stdout or to stderr but not both as you can't guarantee how you'll handle it.

pkg/plugins/builtin/script/script.go
Alkorin: So you prefer to silently ignore and drop errors if the script which should output JSON has some error and write invalid json? 😕

You could let the guess, but is it better if you permit to maintainers to enforce it and return errors if any

pkg/plugins/builtin/script/script.go
w3st3ry: Actually all the tests I wrote (even those not implemented in go_test) who had two steams (stdout/stderr) keep the consistency between common terminal execution and `CombinedOutput()` func. IMO it do the job.
I'll add documentation about that but for the moment we should keep it like this.
If there is issue in the future, I'll do another PR to fix this plugin and SSH.

pkg/plugins/builtin/script/script.go
w3st3ry: Yes I think it could be nice, thanks!

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,159 @@
...
+
+	metadata := Metadata{
+		ExitCode:      fmt.Sprint(exitCode),
+		ProcessState:  pState,
+		Output:        outStr,
+		ExecutionTime: execTime.String(),
+		Error:         metaError,
+	}
+
+	if cfg.AllowExitNonZero && exitCode != 0 {
```
loopfz: !cfg.Allow

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	scriptPath := filepath.Join(utask.FScriptsFolder, cfg.File)
+
+	f, err := os.Stat(scriptPath)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("%s not found in FS: %s", cfg.File, err.Error())
+	}
+
+	if f.Mode()&0111 == 0 {
```
loopfz: f could be nil here? if there was an error other than NotExist

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,159 @@
...
+		if err := os.Chmod(scriptPath, f.Mode()|0111); err != nil {
```
loopfz: chmod from within the binary feels weird: there are possible deployments where I can see this not being ok. it should probably be done in a startup script within the container instead, ie docker entrypoint

w3st3ry: Okey, got it.

pkg/plugins/builtin/script/script.go
w3st3ry: I added a global check of err before.

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+
+	scriptPath := filepath.Join(utask.FScriptsFolder, cfg.File)
+
+	f, err := os.Stat(scriptPath)
+	if err != nil {
+		return fmt.Errorf("can't stat %s: %s", cfg.File, err.Error())
+	}
+
+	if f.Mode()&0111 == 0 {
+		return fmt.Errorf("%s haven't exec permissions", cfg.File)
```
Alkorin:
```suggestion
		return fmt.Errorf("%q is not executable", scriptPath)
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if cfg.File == "" {
+		return errors.New("file is missing")
+	}
+
+	scriptPath := filepath.Join(utask.FScriptsFolder, cfg.File)
+
+	f, err := os.Stat(scriptPath)
+	if err != nil {
+		return fmt.Errorf("can't stat %s: %s", cfg.File, err.Error())
```
Alkorin:
```suggestion
		return fmt.Errorf("can't stat %q: %s", scriptPath, err.Error())
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+	if f.Mode()&0111 == 0 {
+		return fmt.Errorf("%s haven't exec permissions", cfg.File)
+	}
+
+	if cfg.Timeout != "" {
+		if cfg.Timeout[0] == '-' {
+			return errors.New("timeout must be positive")
+		}
+		if _, err := time.ParseDuration(cfg.Timeout); err != nil {
+			return fmt.Errorf("timeout is wrong %s", err.Error())
```
Alkorin:
```suggestion
			return fmt.Errorf("unable to parse duration %q: %s", cfg.Timeout, err.Error())
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+func exec(stepName string, config interface{}, ctx interface{}) (interface{}, interface{}, error) {
+	cfg := config.(*Config)
+
+	var timeout time.Duration
+
+	if cfg.Timeout != "" {
+		timeout, _ = time.ParseDuration(cfg.Timeout)
+	} else {
+		// default is 2 * 1 minute = 2 minutes
+		timeout = time.Duration(2) * time.Minute
```
Alkorin: No need to cast when the number is a constant

```suggestion
		timeout = 2 * time.Minute
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+	lastLine := ""
+
+	for i := len(outputArray) - 1; i >= 0; i-- {
+		if len(outputArray[i]) > 0 {
+			lastLine = outputArray[i]
+			break
+		}
+	}
+
+	if cfg.LastLineNotJSONOutput {
```
Alkorin: Can be moved to line 132, no need to find lastLine if we don't want to parse it :)

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+
+	scriptPath := filepath.Join(utask.FScriptsFolder, cfg.File)
+
+	f, err := os.Stat(scriptPath)
+	if err != nil {
+		return fmt.Errorf("can't stat %s: %s", scriptPath, err.Error())
+	}
+
+	if f.Mode()&0111 == 0 {
+		return fmt.Errorf("%s is not executable", scriptPath)
```
Alkorin: %q is quite useful here, it can help to debug some whitespace issue in config files

```suggestion
		return fmt.Errorf("%q is not executable", scriptPath)
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if cfg.Timeout != "" {
+		if cfg.Timeout[0] == '-' {
+			return errors.New("timeout must be positive")
+		}
+		if _, err := time.ParseDuration(cfg.Timeout); err != nil {
+			return fmt.Errorf("unable to parse duration %s: %s", cfg.Timeout, err.Error())
```
Alkorin: %q is quite useful here, it can help to debug some whitespace issue in config files

```suggestion
			return fmt.Errorf("unable to parse duration %q: %s", cfg.Timeout, err.Error())
```

pkg/plugins/builtin/script/script.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if cfg.File == "" {
+		return errors.New("file is missing")
+	}
+
+	scriptPath := filepath.Join(utask.FScriptsFolder, cfg.File)
+
+	f, err := os.Stat(scriptPath)
+	if err != nil {
+		return fmt.Errorf("can't stat %s: %s", scriptPath, err.Error())
```
Alkorin: %q is quite useful here, it can help to debug some whitespace issue in config files

```suggestion
		return fmt.Errorf("can't stat %q: %s", scriptPath, err.Error())
```

PR #47, opened 2019-12-03
feat: Split payload and metadata from SSH plugin

w3st3ry:
As discussed /w @loopfz, it's better to split JSON payload object and execution metadata.

Signed-off-by: Valentin Pichard <7628998+w3st3ry@users.noreply.github.com>

### DIFF
--- a/pkg/plugins/builtin/ssh/README.md
+++ b/pkg/plugins/builtin/ssh/README.md
@@ -61,31 +61,7 @@ The `Metadata` to fetch informations about plugin execution:
 
 ```json
 {
-  "output": "Connecting...
-    Pseudo-terminal will not be allocated because stdin is not a terminal.
-    Welcome to Ubuntu 19.04 (GNU/Linux 5.0.0-31-generic x86_64)
-
-     * Documentation:  https://help.ubuntu.com
-     * Management:     https://landscape.canonical.com
-     * Support:        https://ubuntu.com/advantage
-
-      System information as of Mon Dec  9 11:03:03 UTC 2019
-
-      System load:  0.0               Processes:           111
-      Usage of /:   23.7% of 9.52GB   Users logged in:     0
-      Memory usage: 18%               IP address for ens3: 10.00.00.01
-      Swap usage:   0%
-
-     * Overheard at KubeCon: \"microk8s.status just blew my mind\".
-
-         https://microk8s.io/docs/commands#microk8s.status
-
-    26 updates can be installed immediately.
-    0 of these updates are security updates.
-
-
-    *** System restart required ***
-    {\"pid\":\"3931\",\"service_name\":\"nginx\",\"service_uptime\":\"1715606\"}",
+  "output": "Connecting...\nWelcome to Ubuntu 19.04 (GNU/Linux ... x86_64)\n[...]\n{\"pid\":\"3931\",\"service_name\":\"nginx\",\"service_uptime\":\"1715606\"}",
   "exit_status": "0",
   "exit_signal": "0",
   "exit_msg": "exited 0"
--- a/pkg/plugins/builtin/ssh/README.md
+++ b/pkg/plugins/builtin/ssh/README.md
@@ -62,8 +62,8 @@ The `Metadata` to fetch informations about plugin execution:
 ```json
 {
   "output": "{\"foo\":\"bar\"}}",
-	"exit_status": "0",
-	"exit_signal": "0",
-	"exit_msg": "exited 0"
+  "exit_status": "0",
+  "exit_signal": "0",
+  "exit_msg": "exited 0"
 }
 ```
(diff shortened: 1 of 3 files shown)


ovh-cds:
CDS Report test#172.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#172.1 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#172.2 ✘  
* Stage 1
  * Test ✘

pkg/plugins/builtin/ssh/README.md
```diff
@@ -42,15 +42,28 @@ action:
...
 ```
 
-## Requirements
+The `Metadata` to fetch informations about plugin execution:
 
-None by default. Ssh credentials should be retrieved from `{{.config.[sshkey]}}` rather than hardcoded on the template.
\ No newline at end of file
+```json
+{
+  "output": "{\"foo\":\"bar\"}}",
```
loopfz: This is not a fitting output. Output = the raw ssh session as it was executed

pkg/plugins/builtin/ssh/README.md
```diff
@@ -42,15 +42,28 @@ action:
...
 ```
 
-The output resulting from this configuration will be:
+## Requirements
+
+None by default. Ssh credentials should be retrieved from `{{.config.[sshkey]}}` rather than hardcoded on the template.
+
+## Note
+
+The plugin returns two objects, the `Payload` of the execution:
```
loopfz: It's worth noting that the format of this payload is set in the **result** field of the step configuration

pkg/plugins/builtin/ssh/ssh.go
```diff
@@ -177,34 +177,36 @@ trap printResultJSON EXIT
...
+	lastLine := outStr[lastNL:]
+	m := map[string]string{}
+
+	err = json.Unmarshal([]byte(lastLine), &m)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for k, v := range m {
+		payload[k] = v
```
loopfz: since there's no need to blend pre-set keys and the actual payload, couldnt we just unmarshal into _payload_ directly ?

pkg/plugins/builtin/ssh/README.md
w3st3ry: Still good now!

pkg/plugins/builtin/ssh/README.md
w3st3ry: fixed.

pkg/plugins/builtin/ssh/ssh.go
w3st3ry: right, thanks!

pkg/plugins/builtin/ssh/README.md
```diff
@@ -42,15 +42,52 @@ action:
...
+  "uptime": "1715606"
 }
 ```
 
-## Requirements
+The `Metadata` to fetch informations about plugin execution:
+
+```json
+{
+  "output": "Connecting...
```
nomonamo: nitpicking: I'm afraid this is not valid json :) you're missing`\n` at the end of each line

pkg/plugins/builtin/ssh/README.md
```diff
@@ -42,15 +42,52 @@ action:
     ssh_key: '{{.config.mySSHKey}}'
 ```
 
-The output resulting from this configuration will be:
+## Requirements
+
+None by default. Ssh credentials should be retrieved from `{{.config.[sshkey]}}` rather than hardcoded on the template.
```
loopfz: I think [sshkey] is unintuitive here: it looks as if it is a specific syntax while it should be clear it's a placeholder name for a config item.

pkg/plugins/builtin/ssh/README.md
loopfz: also it looks a bit weird without the cmds, should we give an example where the script starts with set -x ?

pkg/plugins/builtin/ssh/ssh.go
```diff
@@ -177,34 +177,30 @@ trap printResultJSON EXIT
...
-			return nil, nil, err
-		}
-		for k, v := range m {
-			_, exists := ret[k]
-			// do not overwrite entries already set by plugin (e.g. 'output')
-			if !exists {
-				ret[k] = v
-			}
-		}
+	if lastNL == -1 {
```
loopfz: by returning early you're changing the code paths: exitStatus error wont be triggered here?

w3st3ry: `exitStatus` is already in metadata but you right, it's not consistent /w the previous version.
Fixed.

pkg/plugins/builtin/ssh/README.md
w3st3ry: @loopfz You mean by cmds commands that retrieves pid,service_name... ?

ovh-cds:
CDS Report test#206.0 ✘  
* Stage 1
  * Test ✘

PR #62, opened 2020-01-08
chore: Add a docker build step during CI to validate Dockerfile

rbeuque74:
To prevent some faulty build during releasing process (npm dependencies not matching, Dockerfile invalid), we should run a docker build during CI process.


hack/test-docker.sh
```diff
@@ -2,18 +2,18 @@
 
 export PG_USER="test"
 export PG_PASSWORD="test"
-export PG_PORT="5433"
-export PG_HOST="127.0.0.1"
+export PG_PORT="5432"
 export PG_DATABASENAME="postgres"
 
-PG_DOCKER_ID=$(docker run -e POSTGRES_USER=$PG_USER -e POSTGRES_PASSWORD=$PG_PASSWORD -p$PG_PORT:5432 -d postgres:9.5.3)
```
rbeuque74: Not exposing port to the host will reduce failure if port is already allocated

.travis.yml
```diff
@@ -1,14 +1,18 @@
 os:
   - linux
+dist: bionic
```
w3st3ry: Why do you need this specific dist?

rbeuque74: I pinned it to use the latest distribution (18.04) because default is 16.04.

PR #63, opened 2020-01-09
feat: http plugin improments

fsamin:
**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
HTTP Plugin new features and unit tests

**What is the current behavior?** (You can also link to an open issue here)

**What is the new behavior (if this is a feature change)?**

Improvements on http plugin:
* bearer token authentication
* returns Cookies in metadata

**Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

The `auth_basic` configuration is replaced by `auth`. If think it's more elegant, but we can discuss about it if you prefer to keep it.

**Other information**:

### DIFF
--- a/pkg/plugins/builtin/http/README.md
+++ b/pkg/plugins/builtin/http/README.md
@@ -11,7 +11,7 @@ This plugin permorms an http request.
 | `body` | a string representing the payload to be sent with the request
 | `headers` | a list of headers, represented as objects composed of `name` and `value`
 | `timeout_seconds` | an unsigned int representing a custom HTTP client timeout in seconds
-| `basic_auth` | a single object composed of `user` and `password` to enable HTTP basic auth
+| `auth` | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or `bearer` field to enable Bearer Token Authorization 
 | `deny_redirects` | a boolean representing the policy of redirects
 | `parameters` | a list of HTTP query parameters, represented as objects composed of `key` and `value`
 | `trim_prefix`| prefix in the response that must be removed before unmarshalling (optional)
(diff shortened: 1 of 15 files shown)


pkg/plugins/builtin/http/http.go
```diff
@@ -47,12 +47,41 @@ type Parameter struct {
...
+	Timeout       time.Duration
+	DenyRedirects bool
+}
+
+func defaultHTTPClientFactory(cfg HTTPClientConfig) HTTPClient {
+	c := new(http.Client)
+	c.Timeout = cfg.Timeout
+	if cfg.DenyRedirects {
+
+	}
```
loopfz: wrong scope ?

pkg/plugins/builtin/http/http.go
```diff
@@ -29,7 +29,7 @@ type HTTPConfig struct {
 	Body           string      `json:"body,omitempty"`
 	Headers        []Header    `json:"headers,omitempty"`
 	TimeoutSeconds string      `json:"timeout_seconds,omitempty"`
-	HTTPBasicAuth  BasicAuth   `json:"basic_auth,omitempty"`
+	Auth           Auth        `json:"auth,omitempty"` // FIXME how to manage deprecation of the basic auth attribute ?
```
loopfz: I'm not _too_ afraid or slightly breaking changes right now because the installed user base is small -- that being said, I think long term the best way to deal with such breaking changes is to document them really well in the changelog.
Since the version upgrade is an involved action from the utask end user (updating his Dockerfile FROM, or even re-running the new install script, or make update, or something), displaying this clearly should help fix the affected templates.

pkg/plugins/builtin/http/http.go
fsamin: fixed

pkg/plugins/builtin/http/http.go
fsamin: fine, documentation has been updated

pkg/plugins/builtin/http/README.md
```diff
@@ -30,10 +30,12 @@ action:
     method: POST
     # optional, string as uint16
     timeout_seconds: "5"
-    # optional, object of user and password fields
-    basic_auth:
-      user: {{.config.basicAuth.user}}
-      password: {{.config.basicAuth.password}}
+    # optional, authentication you can user aither basic or bearer auth
```
loopfz:
```suggestion
    # optional, authentication you can user either basic or bearer auth
```

PR #65, opened 2020-01-10
feat: Personalize path prefix for dashboard/editor and API

### LINKED ISSUE #64: Make UI proxypass aware
* **I'm submitting a ...**
  - [x] bug report
  - [x] feature request

* **What is the current behavior?**
When utask is being deployed behind a ProxyPass with a path prefix, UI doesn't behave correctly because it doesn't have the correct base URL for it's own files, and to call API.

* **What is the expected behavior?**
To be able to configure a path prefix while working behind a proxypass

* **Which version of uTask are you using?**
v1.2.0

rbeuque74:
When utask is deployed behind a ProxyPass, UIs need to be aware of where
they are currently running (which path prefix is prepended) & where API
is reachable.

Adding 3 new configuration variable to personalize it:
- dashboard_path_prefix
- dashboard_api_path_prefix
- editor_path_prefix

Dashboard & Editor UIs are served with path_prefix being rewritten at
flight.

Fixes #64

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -8,9 +8,11 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/gin-gonic/gin"
 	"github.com/juju/errors"
 	"github.com/loopfz/gadgeto/iffy"
 	"github.com/loopfz/gadgeto/zesty"
@@ -65,6 +67,9 @@ func TestMain(m *testing.M) {
 
 	srv := api.NewServer()
 	srv.WithAuth(dumbIdentityProvider)
+	srv.SetDashboardPathPrefix("")
+	srv.SetDashboardAPIPathPrefix("")
+	srv.SetEditorPathPrefix("")
 
 	go srv.ListenAndServe()
 	srvx := &http.Server{Addr: fmt.Sprintf(":%d", utask.FPort)}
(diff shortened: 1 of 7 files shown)


Alkorin:
Why _OVH_ prefix?

api/api_test.go
```diff
@@ -530,6 +535,67 @@ func blockedHidden(name string, blocked, hidden bool) tasktemplate.TaskTemplate
...
 
+func Test_staticMiddleware(t *testing.T) {
+	ctx := context.Background()
+
+	ginEngine := gin.Default()
+	ginEngine.
+		Group("/", api.StaticFilePatternReplaceMiddleware("static.go", "___test_suite___")).
+		StaticFS("/", http.Dir("./"))
+
+	srv := &http.Server{Addr: ":12129", Handler: ginEngine}
```
Alkorin: Why ? Can't you do
```
tester := iffy.NewTester(t, ginEngine)
```

You should not have to start a real http Server for testing (and not on an hardcoded TCP port...)

rbeuque74:
> Why _OVH_ prefix?

No real reason, was using this prefix to avoid conflict in files, will use UTASK prefix to make it "company" agnostic

api/api_test.go
rbeuque74: Wasn't sure I could do this, I prefer, thanks for the tip. Fixed

rbeuque74:
Rebased on origin/master

PR #79, opened 2020-01-21
fix: templates: Multiple templates import no longer breaks when depends of each others in sub-task pluggin

### LINKED ISSUE #74: subtask: fail to import multiples templates at first start
* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request

* **What is the current behavior?**
When importing templates at startup, if templates contains some subtasks, the subtask template name needs to already exists in database to be added.
In case of a first deployment, if we have two templates like this:

```yaml
name: hello-world
steps:
    say-hello-world:
        action:
            type: subtask
            configuration:
                template: say-something
                input:
                    who: 'rbeuque'
                    what: 'Hello, world!'
```

```yaml
name: say-something
inputs:
  - name: who
    description: Who
  - name: what
    description: What
steps:
    say-it:
        action:
            type: echo
            configuration:
              output:
                who: '{{.input.who}}'
                what: '{{.input.what}}'
```

hello-world.yaml will be imported first, but will fail with errors, as template contain a subtask, calling template say-something, which haven't been imported yet.

* **What is the expected behavior?**
Check all template name that will be imported during the batch before performing validation.

* **Which version of uTask are you using?**
v1.2.2

rbeuque74:
Fixes #74

### DIFF
--- a/models/tasktemplate/fromdir.go
+++ b/models/tasktemplate/fromdir.go
@@ -9,7 +9,12 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/juju/errors"
 	"github.com/loopfz/gadgeto/zesty"
+	"github.com/ovh/utask/pkg/templateimport"
 	"github.com/sirupsen/logrus"
+)
+
+var (
+	discoveredTemplates []TaskTemplate = []TaskTemplate{}
 )
 
 // LoadFromDir reads yaml-formatted task templates
@@ -31,8 +36,15 @@ func LoadFromDir(dbp zesty.DBProvider, dir string) error {
 		if err := yaml.Unmarshal(tmpl, &tt); err != nil {
 			return fmt.Errorf("failed to unmarshal '%s': '%s'", file.Name(), err)
 		}
-		verb := "Created"
+
 		tt.Normalize()
+
+		discoveredTemplates = append(discoveredTemplates, tt)
+		templateimport.AddTemplate(tt.Name)
+	}
+
+	for _, tt := range discoveredTemplates {
+		verb := "Created"
 		existing, err := LoadFromName(dbp, tt.Name)
 		if err != nil {
 			if !errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


models/tasktemplate/fromdir.go
```diff
@@ -50,5 +60,6 @@ func LoadFromDir(dbp zesty.DBProvider, dir string) error {
 		}
 		logrus.Infof("%s task template '%s'", verb, tt.Name)
 	}
+	templateimport.CleanTemplates()
```
loopfz: Why not leave the list intact?
If you clean it, you cannot call subtask.ValidConfig ever during the runtime, which feels a bit weird/brittle

models/tasktemplate/fromdir.go
```diff
@@ -31,6 +36,11 @@ func LoadFromDir(dbp zesty.DBProvider, dir string) error {
 		if err := yaml.Unmarshal(tmpl, &tt); err != nil {
 			return fmt.Errorf("failed to unmarshal '%s': '%s'", file.Name(), err)
 		}
+		discoveredTemplates = append(discoveredTemplates, tt)
```
loopfz: Why keep a 2nd collection? You could range over templateimport.GetTemplates

rbeuque74: templateimport.GetTemplate only returns a list of template names because I want to avoid cyclic import between templateimport, tasktemplate and the step package.
The second collection contains the full complete tasktemplate items.

models/tasktemplate/fromdir.go
rbeuque74: subtask.ValidConfig at runtime should validate only against template in database, because that's the template we can use. The templateimport collection is just here to registers template names for startup.

I chose to clean up after the loop here to avoid any side effect inside the software of code path that will still use template names for the boot collection instead from the database.

As the Valid func in subtask plugin is looking for template inside database AND the templateimport collection, runtime validation will be okay.

models/tasktemplate/fromdir.go
```diff
@@ -31,6 +36,11 @@ func LoadFromDir(dbp zesty.DBProvider, dir string) error {
...
+		templateimport.AddTemplate(tt.Name)
```
rbeuque74: Need to normalize Name before adding it to collection

PR #82, opened 2020-02-04
fix: tasktemplates: Delete/Archive tasktemplate at import if they no longer exists

rbeuque74:
This will remove tasktemplate from database if the specified template is not in templates folder anymore.
Fixing any problems of remaining templates after being deleted from templates folder.
If tasktemplate still linked to some tasks, then it's hold inside the database, but with flag `hidden` and `blocked` to prevent regular users from using it.


ovh-cds:
CDS Report test#323.1 ✘  
* Stage 1
  * Test ✘

PR #104, opened 2020-04-01
feat: add tags to task

wI2L:

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -342,6 +342,15 @@ func resolve(dbp zesty.DBProvider, res *resolution.Resolution, t *task.Task, deb
 		case s := <-stepChan:
 			s.LastRun = time.Now()
 
+			// Replace task's tags with the tags returned in the step.
+			for k, v := range s.Tags {
+				if v == "" {
+					delete(t.Tags, k)
+				} else {
+					t.Tags[k] = v
+				}
+			}
+
 			// "commit" step back into resolution
 			res.SetStep(s.Name, s)
 			// consolidate its result into live values
@@ -369,7 +378,6 @@ func resolve(dbp zesty.DBProvider, res *resolution.Resolution, t *task.Task, deb
 			if s.IsFinal() && !s.IsChild() {
 				t.StepsDone++
 			}
-
 			// one less step to go
 			expectedMessages--
 			// state change might unlock more steps for execution
@@ -500,11 +508,15 @@ func commit(dbp zesty.DBProvider, res *resolution.Resolution, t *task.Task) erro
 	if err != nil {
 		return err
 	}
-	if err := res.Update(dbp); err != nil {
-		return err
+	if res != nil {
+		if err := res.Update(dbp); err != nil {
+			return err
+		}
 	}
-	if err := t.Update(dbp, false, true); err != nil {
-		return err
+	if t != nil {
+		if err := t.Update(dbp, false, true); err != nil {
+			return err
+		}
 	}
 	return dbp.Commit()
 }
@@ -543,7 +555,7 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res 
 				}
 				// rebuild step dependency tree to include generated loop steps
 				res.BuildStepTree()
-				commit(dbp, res, t)
+				commit(dbp, res, nil)
 				go func() { stepChan <- s }()
 			} else { // regular step
 				s.ResultValidate = jsonschema.Validator(s.Name, s.Schema)
@@ -554,7 +566,7 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res 
 				if s.State != step.StateAfterrunError {
 					s.State = step.StateRunning
 					step.PreRun(s, res.Values, resolutionStateSetter(res, preRunModifiedSteps))
-					commit(dbp, res, t)
+					commit(dbp, res, nil)
 				}
 
 				// run
--- a/engine/step/step.go
+++ b/engine/step/step.go
@@ -103,6 +103,8 @@ type Step struct {
 	Item            interface{}     `json:"item,omitempty"` // "child" step: item value, issued from foreach
 
 	Resources []string `json:"resources"` // resource limits to enforce
+
+	Tags map[string]string `json:"tags"`
 }
 
 // Executor matches an executor type with its required configuration
@@ -252,7 +254,6 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, values *values.Values,
 	}
 
 	go func() {
-
 		limits := uniqueSortedList(st.Resources)
 		for _, limit := range limits {
 			utask.AcquireResource(limit)
@@ -262,7 +263,7 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, values *values.Values,
 		case <-stopRunningSteps:
 			st.State = StateToRetry
 		default:
-			st.Output, st.Metadata, err = runner.Exec(st.Name, baseCfgRaw, config, ctx)
+			st.Output, st.Metadata, st.Tags, err = runner.Exec(st.Name, baseCfgRaw, config, ctx)
 			if baseOutput != nil {
 				if st.Output != nil {
 					marshaled, err := utils.JSONMarshal(st.Output)
--- a/models/task/task.go
+++ b/models/task/task.go
@@ -1,6 +1,7 @@
 package task
 
 import (
+	"encoding/json"
 	"fmt"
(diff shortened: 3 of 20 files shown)


ovh-cds:
CDS Report test#480.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#481.0 ✘  
* Stage 1
  * Test ✘

engine/engine.go
```diff
@@ -342,6 +342,14 @@ func resolve(dbp zesty.DBProvider, res *resolution.Resolution, t *task.Task, deb
 		case s := <-stepChan:
 			s.LastRun = time.Now()
 
+			// Replace task's tags with the tags returned in the step.
```
loopfz, line 345: I think the plugin should return a single map of tags, with the zero value interpreted as a delete instruction.
It may seem counter-intuitive but it's a pretty simple protocol and it lets template writers make add-or-delete decisions with templating instructions.

engine/engine.go
```diff
@@ -543,7 +554,7 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
 				}
 				// rebuild step dependency tree to include generated loop steps
 				res.BuildStepTree()
-				commit(dbp, res, t)
+				commit(dbp, res, nil)
```
loopfz, line 557: I assume there's a reason to not pass t here?

engine/step/step.go
```diff
@@ -103,6 +103,9 @@ type Step struct {
 	Item            interface{}     `json:"item,omitempty"` // "child" step: item value, issued from foreach
 
 	Resources []string `json:"resources"` // resource limits to enforce
+
+	NewTags     map[string]string `json:"new_tags"`
```
loopfz, line 107: Squash both into "ModifiedTags" ?

models/task/task.go
```diff
@@ -62,27 +63,28 @@ type Task struct {
...
+	WatcherUsernames  []string          `json:"watcher_usernames,omitempty" db:"watcher_usernames"`
+	ResolverUsernames []string          `json:"resolver_usernames,omitempty" db:"resolver_usernames"`
+	Created           time.Time         `json:"created" db:"created"`
+	State             string            `json:"state" db:"state"`
+	StepsDone         int               `json:"steps_done" db:"steps_done"`
+	StepsTotal        int               `json:"steps_total" db:"steps_total"`
+	LastActivity      time.Time         `json:"last_activity" db:"last_activity"`
+	Tags              map[string]string `json:"tags,omitempty" db:"tags"`
 
 	CryptKey        []byte `json:"-" db:"crypt_key"` // key for encrypting steps (itself encrypted with master key)
```
loopfz, line 81: Unrelated but this should probably disappear, something to look into @rbeuque74 ?

pkg/plugins/taskplugin/taskplugin.go
```diff
@@ -58,24 +59,29 @@ func (r PluginExecutor) ValidConfig(baseConfig json.RawMessage, config json.RawM
...
-			return nil, nil, errors.Annotate(err, "failed to unmarshal configuration")
+			return nil, nil, nil, nil, errors.Annotate(err, "failed to unmarshal configuration")
 		}
 	}
-
-	return r.execfunc(stepName, cfg, ctx)
+	var newTags map[string]string
+	var tagsToRemove []string
+	if r.tagsFunc != nil {
+		newTags, tagsToRemove = r.tagsFunc(cfg, ctx)
```
loopfz, line 81: The tag func should probably be called after the main exec func, to be able to apply tags based on the result.
To that end, it should have access to the output/metadata/err values.

sql/migrations/002_tags.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+
+ALTER TABLE "task" ADD COLUMN "tags" JSONB;
+ALTER TABLE "task_template" ADD COLUMN "tags" JSONB;
+
+CREATE INDEX "task_tags_idx" ON "task" USING gin (tags jsonb_path_ops);
```
loopfz, line 6: A comment seems worthwhile here to explain the index type choice, e.g. link to doc or similar?

sql/schema.sql
```diff
@@ -59,6 +61,7 @@ CREATE INDEX ON "task"(state);
 CREATE INDEX ON "task"(last_activity DESC);
 CREATE INDEX ON "task"(watcher_usernames);
 CREATE INDEX ON "task"(resolver_usernames);
+CREATE INDEX ON "task" USING gin (tags jsonb_path_ops);
```
loopfz, line 64: same

engine/engine.go
wI2L: @loopfz We've discussed about this with @rbeuque74 and I disagree for a simple reason: a bug/error in the templating system that would result in a variable/input to "render" as an empty string would cause a tag to be deleted by accident. Not only I find it counter-intuitive, but also potentially risky in such case.

loopfz: There are way more severe templating issues, e.g. doing the wrong request in the first place.
This is a non issue imo, and the current implementation makes it very static, which isnt a desirable trait.

engine/engine.go
wI2L: As seen with @rbeuque74 , the task does not need to be commited at this point.

engine/step/step.go
wI2L: Depends on the decision regarding splitting add and remove operations on tags.

pkg/plugins/taskplugin/taskplugin.go
wI2L: I assumed that the config and context inputs were the only source for the tags mutation func, but yes.

sql/migrations/002_tags.sql
wI2L: Ok.

engine/engine.go
loopfz: Ok, that's what I expected but I havent rechecked the flow

pkg/plugins/taskplugin/taskplugin.go
loopfz: Something like that yes. Maybe err too ?

wI2L: Done.

ovh-cds:
CDS Report test#485.0 ✘  
* Stage 1
  * Test ✘

wI2L:
Will squash and sign after review/approval of the last commit.

models/task/task.go
```diff
@@ -500,6 +523,29 @@ func (t *Task) SetState(s string) {
 	t.notifyState(nil)
 }
 
+func (t *Task) SetTags(tags map[string]string, values *values.Values) error {
+	t.Tags = tags
+	if values == nil {
+		return nil
+	}
+	for k, v := range t.Tags {
+		tempk, err := values.Apply(k, nil, "")
```
loopfz, line 532: templating the keys looks like it brings a bunch of problems, eg
```
foo: baz
{{`"foo"`}}: bar
```
Which should have priority? How do we enforce it? It looks pretty loose.
My gut instinct would be to template only on the values.

wI2L: This is backported from the `applyTemplateToMap` function, which template keys and values of a map.
According to the [YAML spec](https://yaml.org/spec/1.2/spec.html#id2764044):

> The content of a mapping node is an unordered set of key: value node pairs, with the restriction that each of the keys is unique.

uTask is using `go-yaml` v2 through its dependency `github.com/ghodss/yaml`, which discard all previous occurences of a duplicate key in a mapping node. See https://github.com/go-yaml/yaml/issues/154.

We can either remove the templating on keys, which puts a limitation on the plugin, or let to the template writers the responsibility of handling such cases.

In your specific example, the first couple `foo:baz` is silently discarded.

engine/step/step.go
```diff
@@ -264,7 +263,7 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, values *values.Values,
 		case <-stopRunningSteps:
 			st.State = StateToRetry
 		default:
-			st.Output, st.Metadata, st.NewTags, st.DeletedTags, err = runner.Exec(st.Name, baseCfgRaw, config, ctx)
+			st.Output, st.Metadata, st.Tags, err = runner.Exec(st.Name, baseCfgRaw, config, ctx)
```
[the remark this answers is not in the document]
wI2L: The empty value on a mapping node can be expressed using both:

- key:
- key: '' or ""

I don't see any particular reason to trim.

rbeuque74:
Thanks !

PR #112, opened 2020-04-07
Fix: engine/values: Better handling of nil and/or empty values

loopfz:
* **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Consistency fixes around templating nil or empty values

SetOutput(step, nil) can be problematic -- it stores a nil element in the values map which may trigger errors in the templating instead of a graceful handling of an empty value (with access to `default` templating function, etc.).
The `field` templating function behaves differently than the dot notation: it returns the litteral `<no value>` which is considered true in templating pipelines, thus not compatible with functions like `default`.

SetOutput and other functions setting step data now avoid setting nil values, replacing them with an empty map.
The field templating function now returns a reflect.Value similar to builtin functions inside text template. This allows us to return a zero-value reflect.Value when not found, which properly evals to false inside templating pipelines (`default` function ok), and prints to `<no value>` outside of templating pipelines.

`fieldjson` and `jsonmarshal` were also problematic. Since they are redundant with built-in sprig functions (`toJson`, ... http://masterminds.github.io/sprig/defaults.html ), they got removed.

* **Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,6 @@ The following templating functions are available:
 |**`Golang`** | Builtin functions from Golang text template  | [Doc](https://golang.org/pkg/text/template/#hdr-Actions)
 |**`Sprig`** | Extended set of functions from the Sprig project  | [Doc](https://masterminds.github.io/sprig/)
 |**`field`** | Equivalent to the dot notation, for entries with forbidden characters | ``{{field `config` `foo.bar`}}``
-|**`jsonmarshal`** | Used as part of a templating pipeline, outputs a JSON representation | ``{{.step.foo.output.bar \| jsonmarshal}}``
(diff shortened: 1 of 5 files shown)


ovh-cds:
CDS Report test#500.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#501.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#502.0 ✘  
* Stage 1
  * Test ✘

PR #118, opened 2020-04-08
feat: uTask API server is YAML capable

rbeuque74:


rbeuque74:
Build is failing because no vendor as been generated, we should merge #116 before.

PR #127, opened 2020-04-15
fix: engine: tasks could stay in inconsistent state when fails to validate templates

rbeuque74:
When a task fails to validate requirements of a template, task is set to
BLOCKED state, whereas resolution stayed in RUNNING state, making it
impossible for admins to manage the task anymore.
Switching resolution to a CRASHED state will allow manipulation by
admins.

Fixing a bug

Task that failed to validate a template was in BLOCKED state with resolution in RUNNING state, couldn't be used in the API anymore.

Resolution in CRASHED state

No

* **Other information**:

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -190,7 +190,7 @@ func (e Engine) launchResolution(publicID string, async bool) (*resolution.Resol
 	}
 
 	// check/update states for all concerned objects
-	res, t, err := initialize(dbp, publicID)
+	res, t, err := initialize(dbp, publicID, debugLogger)
 	if err != nil {
 		debugLogger.Debugf("Engine: Resolve() %s initialize error: %s", publicID, err)
 		return nil, err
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -294,7 +294,11 @@ func initialize(dbp zesty.DBProvider, publicID string) (*resolution.Resolution,
 		}
 		// task validation error
 		// -> possible race condition with updated template
-		// put task on hold
+		// put task + resolution on hold
+		res.SetState(resolution.StateCrashed)
```
loopfz, line 298: CRASHED seems weird to me.
Looking at the current collectors I don't see any of them auto picking CRASHED resolutions, but it could certainly happen at some point.
This on the other hand doesnt seem very retryable, there's a mismatch.
Since we're putting the task on BLOCKED, I'd lean towards a flavor of BLOCKED_XXX for the resolution itself. BLOCKED_FATAL or BLOCKED_BADREQUEST come to mind.

rbeuque74: What about ?
```suggestion
		res.SetState(resolution.StateBlockedToCheck)
```

ovh-cds:
CDS Report test#575.0 ✘  
* Stage 1
  * Test ✘

PR #129, opened 2020-04-16
feat: add IN and NOTIN to step condition operators

wI2L:
These operators allows to check that a value is present or not
in a list of values. The allowed/disabled values are represented
as a comma-separated string.

Feature.

A value cannot be checked against a list of potential candidates in a step condition.

This permits to assert that a value is present or not in a list of acceptable values.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,13 +177,14 @@ A user can be allowed to resolve a task in three ways:
 - `.iterator.foo`: field `foo` from the iterator in a loop (see `foreach` steps below)
 
 The following templating functions are available:
-|Name|Description|Reference
-|---|---|---
-|**`Golang`** | Builtin functions from Golang text template  | [Doc](https://golang.org/pkg/text/template/#hdr-Actions)
-|**`Sprig`** | Extended set of functions from the Sprig project  | [Doc](https://masterminds.github.io/sprig/)
-|**`field`** | Equivalent to the dot notation, for entries with forbidden characters | ``{{field `config` `foo.bar`}}``
-|**`eval`** | Evaluates the value of a template variable | ``{{eval `var1`}}``
-|**`evalCache`** | Evaluates the value of a template variable, and cache for future usage (to avoid further computation) | ``{{evalCache `var1`}}``
+
+| Name            | Description                                                                                           | Reference                                                |
+| --------------- | ----------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
+| **`Golang`**    | Builtin functions from Golang text template                                                           | [Doc](https://golang.org/pkg/text/template/#hdr-Actions) |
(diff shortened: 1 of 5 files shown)


engine/step/condition.go
```diff
@@ -113,6 +123,16 @@ func (a *Assert) Valid() error {
 	return nil
 }
 
+func matchList(valStr, expStr string) bool {
+	values := strings.Split(expStr, ",")
+	for _, v := range values {
+		if valStr == v {
```
loopfz, line 129: I'm always paranoid about whitespaces messing up comparisons.
Should we TrimSpace(v) ?

loopfz: If complexity is a problem, we could have a Prepare() step on resolution load

engine/templates_tests/stepCondition.yaml
```diff
@@ -76,3 +76,31 @@ steps:
...
+        custom_states: [LIST_MATCH,LIST_NOMATCH]
+        action:
+            type: echo
+            configuration:
+                output:
+                    items: 'foo,bar'
+        conditions:
+            - type: check
+              if:
+                  - value: '{{.step.this.output.items}}'
```
loopfz, line 91: the condition is backwards, isnt it?

foo IS IN foo,bar

wI2L: Indeed, it doesn't match the code. Fixing.

loopfz:
Looks good overall, 2 issues to address imo:
- whitespace trim (see earlier comment)
- update doc to cover the new operators, mentionning specifically the order of operands (expected vs value)

README.md
```diff
@@ -293,6 +294,21 @@ steps:
...
+A condition can use one of the following operators:
+- `EQ`: equal
+- `NE`: not equal
+- `GT`: greater than
+- `LT`: less than
+- `GE`: greater or equal
+- `LE`: less than or equal
+- `REGEXP`: match a regexp
+- `IN`: found in a list of values
+- `NOTIN`: doesn't found in a list of values
```
loopfz, line 308:
```suggestion
- `NOTIN`: not found in a list of values
```

engine/templates_tests/stepCondition.yaml
```diff
@@ -76,3 +76,31 @@ steps:
               then:
                   this: FOO
               message: changed state matched correctly
+    stepFour:
```
rbeuque74, line 79: Maybe could you patch engine_test.go to make sure that the expected custom_state is correctly applied, so we can validate that your code is working.

ovh-cds:
CDS Report test#600.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#648.0 ✘  
* Stage 1
  * Test ✘

PR #135, opened 2020-04-24
functions: new way to describe utask executors using other executor, with yaml

ybriffa:
* **What kind of change does this PR introduce?**
It introduces a new feature: functions.

There is no functions

Functions allows to declare a directory with function that can be re-used recursively in templates to hold piece of config and abstract behavior

no

* **Other information**:


ovh-cds:
CDS Report test#609.0 ✘  
* Stage 1
  * Test ✘

engine/step/step.go
```diff
@@ -263,14 +289,14 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, values *values.Values,
 			st.State = StateToRetry
 		default:
 			st.Output, st.Metadata, st.Tags, err = runner.Exec(st.Name, baseCfgRaw, config, ctx)
-			if baseOutput != nil {
-				if st.Output != nil {
+			if len(baseOutputs) > 0 && st.Output != nil {
```
rbeuque74, line 292: We could have some `baseOutputs` without having `output` from the actual runner. In that case, we should apply `baseOutputs` anyway

engine/step/step.go
```diff
@@ -169,62 +170,87 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, values *values.Values,
...
-		st.State = StateFatalError
-		st.Error = err.Error()
-		go noopStep(st, stepChan)
-		return
+		// Check if we have a function as runner or not. If not, we do not need to go further in the resolution
+		functionRunner, ok := runner.(*functions.Function)
+		if !ok {
+			break
+		}
+		var functionIntput map[string]interface{}
```
rbeuque74, line 244:
```suggestion
		var functionInput map[string]interface{}
```

rbeuque74: Still present

docker-compose.yaml
```diff
@@ -16,6 +16,7 @@ services:
     volumes:
       - "./templates:/app/templates:ro"
       - "./scripts:/app/scripts:ro"
+      - "./services:/app/services:ro"
```
rbeuque74, line 19:
```suggestion
      - "./functions:/app/functions:ro"
```

PR #136, opened 2020-04-27
Multiple state dependencies support

rbeuque74:
Feature + typo + docs update

Step dependency was only supporting one state to be matched

We can now declare multiple states for a step dependency. Step will run if one of the given states is matching, otherwise, will be prune as the current behaviour.

No. Just refactoring of engine.go dependency handling.

* **Other information**:


ovh-cds:
CDS Report test#619.0 ✘  
* Stage 1
  * Test ✘

README.md
```diff
@@ -410,6 +418,21 @@ Browse [builtin actions](./pkg/plugins/builtin)
 |**`ping`**    | Send a ping to an hostname *Warn: This plugin will keep running until the count is done* | [Access plugin doc](./pkg/plugins/builtin/ping/README.md)
 |**`script`**    | Execute a script under `scripts` folder | [Access plugin doc](./pkg/plugins/builtin/script/README.md)
 
+#### Dependencies <a name="dependencies"></a>
+
+Dependencies can be declared on a step, to indicate what are the requirements to be met before the step can actually run. A step can have multiple dependencies, which will have to be all matched before the step can be running.
```
loopfz, line 423:
```suggestion
Dependencies can be declared on a step, to indicate what requirements should be met before the step can actually run. A step can have multiple dependencies, which will all have to be met before the step can start running.
```

README.md
```diff
@@ -410,6 +418,21 @@ Browse [builtin actions](./pkg/plugins/builtin)
...
+
+A dependency can be qualified with a step's state. Two kinds of step states exists: builtins and customs. Builtins are provided by uTask and include: `TODO`, `RUNNING`, `DONE`, `CLIENT_ERROR`, `SERVER_ERROR`, `FATAL_ERROR`, `CRASHED`, `PRUNE`, `TO_RETRY`, `AFTERRUN_ERROR`. Each step can also have some custom states, declared within the step using `custom_states`. Dependency can be all custom states declared, however, it can only use the `DONE` builtin state, as it's the only one considered as "final state". Final state represents states that can no longer be run by uTask engine ; all custom states are considered as "final states".
```
loopfz, line 425:
```suggestion
A dependency can be qualified with a step's state (`stepX:stateY`, it depends on stepX, finishing in stateY). If omitted, then `DONE` is assumed.

There are two different kinds of states: builtin and custom. Builtin states are provided by uTask and include: `TODO`, `RUNNING`, `DONE`, `CLIENT_ERROR`, `SERVER_ERROR`, `FATAL_ERROR`, `CRASHED`, `PRUNE`, `TO_RETRY`, `AFTERRUN_ERROR`. Additionally,  a step can define custom states via its `custom_states` field. These custom states provide a way for the step to express that it ran successfully, but the result may be different from the normal expected case (e.g. a custom state `NOT_FOUND` would let the rest of the workflow proceed, but may trigger additional provisioning steps).

A dependency (`stepX:stateY`) can be on any of `stepX`'s custom states, along with `DONE` (builtin). These are all considered final (uTask will not touch that step anymore, it has been run to completion). Conversely, other builtin states (`CLIENT_ERROR`, ...) may not be used in a dependency, since those imply a transient state and the uTask engine still has work to do on these.

If you wish to declare a dependency on something normally considered as a `CLIENT_ERROR` (e.g. GET HTTP returns a 404), you can write a `check` condition to inspect your step result, and change it to a custom state instead (meaning an alternative termination, see the `NOT_FOUND` [example](https://github.com/ovh/utask#steps))
```

README.md
```diff
@@ -410,6 +418,21 @@ Browse [builtin actions](./pkg/plugins/builtin)
...
+
+It is possible that a dependency will never match the expected state. For example, `step1` is in `DONE` state, and `step2` has a dependency declared as `step1:NOT_FOUND`: it means that `step2` requires that `step1` finishes its execution with state `NOT_FOUND`. In that case, `step2` will never be allowed to run, as `step1` finished with `DONE` state ; `step2` state will be changed to `PRUNE`, meaning that the execution of this step hasn't be selected, and won't be run. All the steps which have declared `step2` as a dependency will also be pruned.
```
loopfz, line 427:
```suggestion
It is possible that a dependency will never match the expected state. For example, `step1` is in `DONE` state, and `step2` has a dependency declared as `step1:NOT_FOUND`: it means that `step2` requires that `step1` finishes its execution with state `NOT_FOUND`. In that case, `step2` will never be allowed to run, as `step1` finished with  state `DONE`. To remedy this, uTask will remove `step2` from the workflow by setting its state to the special state `PRUNE`. Any further step depending on `step2` will also be pruned, removing entire alternative execution branches. This allows crossroads patterns, where a step may be followed by two mutually exclusive branches (one for `DONE`, one for `ALTERNATE_STATE_XXX`). (Note: `PRUNE` may also be used in conditions to manually eliminate entire branches of execution)
```

README.md
```diff
@@ -410,6 +418,21 @@ Browse [builtin actions](./pkg/plugins/builtin)
...
+
+Another qualifier that can be use as a step state is `ANY`. `ANY` indicates that dependency will match all the "final states" of the step. It's also the only way for a `step2` with a dependency on `step1` which is in the `PRUNE` state to not be pruned as well: if the dependency has been declared to state `ANY`, the prune chain will be interrupted.
```
loopfz, line 429:
```suggestion
A special qualifier that can be used as a dependency state is `ANY` (`stepX:ANY`). `ANY` matches all custom states and `DONE`, and it also does not get `PRUNE`'d recursively if `stepX` is set to `PRUNE`. This is used mostly for sequencing, either when the actual result of the step does not matter, but its timing does; or to reconcile mutually exclusive branches in a diamond pattern (using e.g. the `coalesce` templating function to mix optional step results).
```

PR #142, opened 2020-05-04
feature: can unmarshal echo output before returning

ybriffa:
A feature to unmarshal echo content before returning

We cannot unmarshal

We can unmarshal

No

* **Other information**:

### DIFF
--- a/pkg/plugins/builtin/echo/README.md
+++ b/pkg/plugins/builtin/echo/README.md
@@ -8,7 +8,7 @@ This plugin returns the output defined in its configuration, without performing 
 |---|---
 | `output` | an object with the complete output of the step
 | `metadata` | an object containing the metadata returned by the step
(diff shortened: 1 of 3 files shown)


pkg/plugins/builtin/echo/README.md
```diff
@@ -8,6 +8,7 @@ This plugin returns the output defined in its configuration, without performing
 |---|---
 | `output` | an object with the complete output of the step
 | `metadata` | an object containing the metadata returned by the step
+| `unmarshal` | defines whether unmarshal the output if it's a string or byte array before returning
```
loopfz, line 11:
```suggestion
| `unmarshal` | if `true`, then `output` is expected to be a single string containing valid JSON; this is useful for generating an object via templating, since uTask can normally only apply templating on strings
```

PR #147, opened 2020-05-20
Dev/ybriffa/pre hook

ybriffa:
It introduces the pre hooks

No prehooks

We can use prehooks in templates steps

No

* **Other information**:


ovh-cds:
CDS Report test#690.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#704.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#705.0 ✘  
* Stage 1
  * Test ✘

ovh-cds:
CDS Report test#721.0 ✘  
* Stage 1
  * Test ✘

ybriffa:
> Looks good to me overall.
> General comments:
> 
>     * the README could also be edited on the Value templating part (https://github.com/ovh/utask#value-templating) to add `.pre_hook` and `.function_args`to add these new values to the global templating values array, even if it's detailed in the according sections.
> 
>     * I'm not sure about a specific behaviour with chained functions that have pre_hooks: pre_hooks will be overrided to keep the deepest one. It could create a weird expectation. Wondering if we should not prevent that at template parsing time.
> 
> 
> Example:
> templates/foo.yaml
> 
> ```yaml
> name: foo
> description: foo
> title_format: foo
> auto_runnable: true
> 
> steps:
>   funcecho:
>     pre_hook:
>       type: echo
>       configuration:
>         output:
>           zozo: foobar
>     description: funcecho
>     idempotent: true
>     action:
>       type: funcecho-with-otp
>       configuration:
>         foo1: DJDJDJDJ
>         foo3: "{{.pre_hook.output.zozo}}"
> ```
> 
> functions/funcecho-with-otp.yaml
> 
> ```yaml
> name: funcecho-with-otp
> pre_hook:
>   type: http
>   configuration:
>     method: "GET"
>     url: https://eu.api.ovh.com/1.0/auth/time
> action:
>   type: funcecho
>   configuration:
>     hellootp: worldotp
>     foo1: "{{.function_args.foo1}}"
>     foo2: "{{.pre_hook.output}}"
>     foo3: "{{.function_args.foo3}}"
> ```
> 
> functions/funcecho.yaml
> 
> ```yaml
> name: funcecho
> action:
>   type: echo
>   configuration:
>     output:
>       hello: world
>       foo1: "{{.function_args.foo1}}"
>       foo2: "{{.function_args.foo2}}"
>       foo3: "{{.function_args.foo3}}"
> ```
> 
> Output being:
> 
> ```js
> {
>     "foo1": "DJDJDJDJ",
>     "foo2": "1591289731",
>     "foo3": "{prehook{.pre_hook.output.zozo}prehook}",
>     "hello": "world"
> }
> ```
> 
> I guess this is what we can expect, but we might be able to warn at template & function loading that the parent function/action will override the pre_hook and that it should not use `.pre_hook` inside its templating. What do you think ?

README updated.
About the prehook override: we discussed about just taking the last one and use it like this. I see that it can lead to missunderstanding, so I added a check to do not override them at all atm. let's see in the future if we encounter this case and we'll take the good decision in consequences ? I have the feeling that we cannot take the good decision until we face a real use case

pkg/plugins/plugins.go
```diff
@@ -53,10 +54,15 @@ type InitializerPlugin interface {
 // InitializersFromFolder loads initialization plugins compiled as .so files
 // from a folder, runs them on a received pointer to a Service
 func InitializersFromFolder(path string, service *Service) error {
-	return loadPlugins(path, func(fileName string, p plugin.Symbol) error {
-		plug, ok := p.(InitializerPlugin)
+	return loadPlugins(path, func(fileName string, pluginSymbol plugin.Symbol) error {
```
rbeuque74, line 57: This code is already merged on master, and should not be contained in the pull-request.
You should rebase on master instead of actual origin/dev/ybriffa/master

engine/values/values.go
```diff
@@ -35,6 +37,7 @@ const (
 type Values struct {
 	m       map[string]interface{}
 	funcMap map[string]interface{}
+	delims  [2]string
```
rbeuque74, line 40: Not needed anymore since we are not resolving templating multiple time

engine/values/values.go
```diff
@@ -68,6 +71,21 @@ func NewValues() *Values {
...
+func (v *Values) Clone() *Values {
+	n := NewValues()
+	for key, value := range v.m {
+		n.m[key] = deepcopy.Copy(value)
+	}
+	return n
+}
+
+// SetDelims changes the delimiter on which resolve the templates variables
+func (v *Values) SetDelims(left, right string) *Values {
```
rbeuque74, line 84: Not needed anymore since we are not resolving templating multiple time

engine/step/step.go
```diff
@@ -138,159 +139,241 @@ func uniqueSortedList(s []string) []string {
...
+	runner           Runner
+	ctx              interface{}
+	stopRunningSteps <-chan struct{}
+}
 
-	// Step already ran, directly going to afterrun process
-	if st.State == StateAfterrunError {
-		go noopStep(st, stepChan)
-		return
+func (e *execution) applyValues(values *values.Values, item interface{}, name string) error {
```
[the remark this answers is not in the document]
ybriffa: one of the functions I told you the next PR would delete :D anyway cleaning up

engine/values/values.go
```diff
@@ -98,6 +116,14 @@ func (v *Values) SetOutput(stepName string, value interface{}) {
 	v.setStepData(stepName, OutputKey, value)
 }
 
+// SetPreHook stores a step's prehook in Values
+func (v *Values) SetPreHook(output, metadata interface{}) {
+	v.m[PreHookKey] = map[string]interface{}{
```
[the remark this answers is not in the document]
ybriffa: I don't think so. We initialise it because in the other function we directly access the map without checking and instantiating the map if not exist. This function SetPreHook instantiate every time the map and we have no access to the keys (except in text/template that manage it very well by itself)

PR #151, opened 2020-05-25
Add POST resolution reinstitute endpoint

zhouzhuojie:

This PR adds a feature so that we can combine `pause`, `update`, and `run` into one api: `reinstitute`. This is useful for the webhook callback action or workflow. 

Example of usage of this endpoint with resolver input and condition:
```yaml

resolver_inputs:                       
- name: callback_state       
  description: callback state
  default: pending                     

steps:
  send_http:
    action:                                                                     
      type: http                                                                
      configuration:                                                            
        url: xxx
        method: xxx                                                          
        body: xxx
        # This example is using headers, you can also use path parameter or body.
        # Depends on the remote callback pattern, we just need to tell them how to send the callback to utask
        # Notice that we need .task.resolution_id, which is also added in this PR
        # Remote server is likely to send request like
        #     curl -H 'Content-Type: application/json' -d '{"resolver_input": {"callback_state": "done"}}' 'http://utask:8081/resolution/{{.task.resolution_id}}/reinstitute'
        headers:                                                                
        - name: x-callback-url                                                  
          value: 'http://utask:8081/resolution/{{.task.resolution_id}}/reinstitute'

  wait_for_callback:
    action:                                                            
      type: echo                                                       
      configuration:                                                   
        output:                                                        
          state: '{{.resolver_input.callback_state}}'        
    conditions:                                                        
    - type: check                                                      
      if:                                                              
      - value: '{{.resolver_input.callback_state}}'          
        operator: NE                                                   
        expected: 'done'                                               
      then:                                                            
        this: TO_RETRY                                                 
      message: 'pending and waiting for callback'                                  
```

- Currently we need to send 3 APIs
- Currently resolution_id is not available in the values

- Added new `/resolution/:resolution_id/reinstitute` api
- Added `resolution_id` in the templating value context

No. See unit tests.

* **Other information**:

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -270,6 +270,179 @@ func TestPasswordInput(t *testing.T) {
 	tester.Run()
 }
 
+func TestResolution(t *testing.T) {
+	t.Run("simple run resolution workflow", func(t *testing.T) {
+		tester := iffy.NewTester(t, hdl)
+		defer tester.Run()
+
+		dbp, err := zesty.NewDBProvider(utask.DBName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		tmpl := templateWithPasswordInput()
+
+		_, err = tasktemplate.LoadFromName(dbp, tmpl.Name)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				t.Fatal(err)
+			}
+			if err := dbp.DB().Insert(&tmpl); err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		tester.AddCall("newTask", http.MethodPost, "/task", `{"template_name":"input-password","input":{"verysecret":"abracadabra"}}`).
+			Headers(regularHeaders).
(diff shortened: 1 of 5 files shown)


loopfz:
Interesting idea.
I'm not sure it's viable from a security standpoint though?

If you check the API RBAC you'll see that the `resolution update` action is reserved to `administrator` users, for one very good reason: updating the resolution allows you to change *anything* (necessary sometime to fix stuck tasks, edge cases, etc. and let them finish).

As such, you can change the destination of a call and leak credentials, change the order of steps, remove critical consistency check steps, etc.. all of which corrupt the task's integrity and the instance's security.

This doesnt seem desirable for a callback process; what do you think ?

zhouzhuojie:
Thanks for the init plugin fix from https://github.com/ovh/utask/pull/152, I think I can add webhooks now from an init plugin. I moved the `resolution_id` to a new PR https://github.com/ovh/utask/pull/154.

So going to close this PR.

PR #160, opened 2020-06-25
fix: recursive templating & foreach steps indefinitely blocking the resolution

rbeuque74:
Bug fixes

\- foreach steps that falls into a FATAL_ERROR state are never picked up
\- failure to template recursively in a map was not catched up correctly

\- fixed
\- fixed

No

* **Other information**:


engine/engine.go
```diff
@@ -585,7 +585,18 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
...
-					continue
+					// if the ForEach task is in another state, we do nothing, but need to return the step in the stepChan
+					// otherwise the task will wait infinitely for this step
+				}
+
+				// after expanding or contracting, need to check if step didn't turned into a bad state (FatalError, ServerError, ...) during the operation
+				// in that case, it should be discarded from being looked-up during next runAvailableSteps (as we did expanded++)
+				switch s.State {
+				case step.StateTODO, step.StateToRetry, step.StateExpanded:
+					// thoses states could be lead the step to be reconsidered on the next runAvailableSteps if the childrens runs fast enough (and not waiting for the next resolution.Run)
```
loopfz, line 596:
```suggestion
					// those states could lead the step to be reconsidered on the next runAvailableSteps if the children run fast enough (and not have to wait for the next resolution.Run)
```

engine/engine.go
```diff
@@ -585,7 +585,18 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
...
+				default:
+					// others states are errors, we should stop considering this step as eligible during this run
```
loopfz, line 598:
```suggestion
					// other states are errors, we should stop considering this step as eligible during this run
```

engine/engine.go
```diff
@@ -585,7 +585,18 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
 					contractStep(s, res)
 					res.SetStepState(s.Name, step.StateDone)
 				default:
-					continue
+					// if the ForEach task is in another state, we do nothing, but need to return the step in the stepChan
+					// otherwise the task will wait infinitely for this step
+				}
+
+				// after expanding or contracting, need to check if step didn't turned into a bad state (FatalError, ServerError, ...) during the operation
```
loopfz, line 592:
```suggestion
				// after expanding or contracting, need to check if the step didn't turn into a bad state (FatalError, ServerError, ...) during the operation
```

engine/engine.go
```diff
@@ -585,7 +585,18 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
...
+				// in that case, it should be discarded from being looked-up during next runAvailableSteps (as we did expanded++)
```
loopfz, line 593:
```suggestion
				// if so, it should be ignored during next runAvailableSteps (as we did expanded++)
```

Makefile
```diff
@@ -3,7 +3,9 @@ BINARY			= utask
 MAIN_LOCATION	= ./cmd
 
 TEST_LOCATION	= ./...
-TEST_CMD		= go test -count=1 -v -cover -p 1 ${TEST_LOCATION}
+# timeout for go test is here to prevent test running infinitely if one runResolution lead to a step that never recovers (missing push on a stepChan)
```
loopfz, line 6:
```suggestion
# timeout for go test is here to prevent tests running infinitely if one runResolution leads to a step that never recovers (e.g. missing push on a stepChan)
```

PR #164, opened 2020-07-11
feat: add fromJson and mustFromJson to values func map

wI2L:

Feature: This adds two new functions `fromJson` and `mustFromJson` to the func map of the `Values` object.

Currently, one can use the `toJson` function of the Sprig project to encode a value to JSON, but the inverse is not available.

A JSON document can be decoded into a data structure, e.g ``{{eval `varReturningJSON` | fromJson}}``, or ``{{fromJson `{"a":"b"}`}}``.

New feature, no breaking changes.

* **Other information**:

Similar feature request is still on hold for Sprig: https://github.com/Masterminds/sprig/pull/230


ovh-cds:
CDS Report test#763.0 ✘  
* Stage 1
  * Test ✘

engine/values/values.go
```diff
@@ -374,6 +378,19 @@ func (v *Values) varEval(varName string) (interface{}, error) {
...
 }
 
+// fromJSON decodes JSON into a structured value, ignoring errors.
+func (v *Values) fromJSON(s string) (interface{}, error) {
+	output, _ := v.mustFromJSON(s)
+	return output, nil
+}
+
+// mustFromJSON decodes JSON into a structured value, returning errors.
+func (v *Values) mustFromJSON(s string) (interface{}, error) {
```
loopfz, line 388: How about returning a (possibly zero value) reflect.Value instead of interface?
See `field` implementation, it gets properly considered as a zero value for eg coalesce, default, '<no value>' replacement by the text tempmate engine, etc

README.md
```diff
@@ -182,13 +182,15 @@ A user can be allowed to resolve a task in three ways:
...
+| **`Sprig`**        | Extended set of functions from the Sprig project                                                                           | [Doc](https://masterminds.github.io/sprig/)              |
+| **`field`**        | Equivalent to the dot notation, for entries with forbidden characters                                                      | ``{{field `config` `foo.bar`}}``                         |
+| **`eval`**         | Evaluates the value of a template variable                                                                                 | ``{{eval `var1`}}``                                      |
+| **`evalCache`**    | Evaluates the value of a template variable, and cache for future usage (to avoid further computation)                      | ``{{evalCache `var1`}}``                                 |
+| **`fromJson`**     | Decodes a JSON document into a structure. If the input cannot be decoded as JSON, the function will return an empty string | ``{{fromJson `{"a":"b"}`}}``                             |
```
loopfz, line 192: I think it'd be interesting to describe a common use case for this here

engine/values/values.go
wI2L: Good idea. What should we consider as "zero-value" in the context of a JSON-value tho ?
`reflect.ValueOf(bill)` will return the zero `Value`, which is great, but what about empty string, zero numbers, and so on ?

README.md
wI2L: Yes. I guess, the most common use case we could document is to return a JSON stringified data structure from a JavaScript expression (object, array), and use one of its members in the template.
Example: ``{{(eval `myExpression` | fromJson).myArr}}`` or ``{{(eval `myExpression` | fromJson).myObj}}``.

wI2L:
One issue I couldn't figure how to circumvent yet, is the fact that a value extracted from a data-structure decode with the introduced functions will always be converted to the Golang textual representation of their type.

See the following test case `TestJSONParsing`:
```go
  assert.Equal(t, "utask", output["a"])
  assert.Equal(t, "666", output["b"])
  assert.Equal(t, "map[k:v]", output["c"])
  assert.Equal(t, "[1 2 3]", output["d"])
```

We are bound by the YAML spec, since we MUST quote our blocks in the template:

```yaml
  a: '{{(eval `rawJSON` | fromJson).a}}'
  b: '{{(eval `rawJSON` | mustFromJson).b}}'
  c: '{{(eval `rawJSON` | mustFromJson).c}}'
  d: '{{(eval `rawJSON` | fromJson).d}}'
``` 

The idea is to be able to pass a map or slice as the input of a subtask without having the value converted to a string.

```yaml
  myStep:
    action:
      type: subtask
      configuration:
        template: my-template
        input:
          foobar: '{{eval `myVariable` | fromJson}}' # <--

```

The field `input` is a `map[string]interface{}`, how do we pass a Go slice without the string conversion when the `myVariable` expression evaluation returns a JSON array ?

wI2L:
I think we can't do more about about the issue I raised in my previous comment https://github.com/ovh/utask/pull/164#issuecomment-657083354. The implementation is good enough for now, we can still leverage the new `fromJSON` functions for useful cases.

@rbeuque74 A second review would be nice, I'll address the minor changes suggested by @loopfz.

rbeuque74 (approved):

loopfz (approved):

PR #166, opened 2020-07-14
subtask: auto-resume parent task when the subtask finished running

rbeuque74:
Feature

In case we have a template running some subtasks, we should wake up the
parent task after one of the subtasks has reached completion

No

* **Other information**:
No


rbeuque74:
Added a validation of tags to prevent non-admins users to use this tag, in order to re-run a resolution that doesn't belongs to them.

loopfz:
Fair enough, lgtm

PR #167, opened 2020-07-17
feat: add custom middlewares for init plugins

zhouzhuojie:
A new feature for init plugin to add custom middlewares.

Currently there's no easy way to add custom middlewares.

A new function `WithCustomMiddlewares`.

No.

* **Other information**:

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -58,7 +58,7 @@ func (s *Server) WithAuth(authProvider func(*http.Request) (string, error)) {
 // WithCustomMiddlewares sets an array of customized gin middlewares.
 // It helps for init plugins to include these customized middlewares in the api server
 func (s *Server) WithCustomMiddlewares(customMiddlewares ...gin.HandlerFunc) {
-	s.customMiddlewares = customMiddlewares
+	s.customMiddlewares = append(s.customerMiddlewares, customMiddlewares...)
 }
 
 // SetDashboardPathPrefix configures the custom path prefix for dashboard static files hosting.
--- a/api/server.go
+++ b/api/server.go
@@ -58,7 +58,7 @@ func (s *Server) WithAuth(authProvider func(*http.Request) (string, error)) {
 // WithCustomMiddlewares sets an array of customized gin middlewares.
 // It helps for init plugins to include these customized middlewares in the api server
 func (s *Server) WithCustomMiddlewares(customMiddlewares ...gin.HandlerFunc) {
-	s.customMiddlewares = append(s.customerMiddlewares, customMiddlewares...)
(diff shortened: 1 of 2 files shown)


api/server.go
```diff
@@ -178,6 +185,7 @@ func (s *Server) build(ctx context.Context) {
 			},
 		})
 
+		router.Use(s.customMiddlewares...)
```
zhouzhuojie, line 188: Not sure if there's better place to inject custom middlewares, for example, another option is to put it right after the line
```
ginEngine := gin.Default()
ginEngine.Use(s.customMiddlewares...)
```

Since the original code has ajaxHeadersMiddleware and errorLogMiddleware, I picked a location close to that.

loopfz:
Looking into this now!

loopfz:
Overall it looks like a fine idea, a lot of flexibility is added and the only notable downside is the added adherence to gin in initPlugins. I think that's a worthy tradeoff.
@rbeuque74 thoughts ?

api/server.go
```diff
@@ -54,6 +55,12 @@ func (s *Server) WithAuth(authProvider func(*http.Request) (string, error)) {
 	}
 }
 
+// WithCustomMiddlewares sets an array of customized gin middlewares.
+// It helps for init plugins to include these customized middlewares in the api server
+func (s *Server) WithCustomMiddlewares(customMiddlewares ...gin.HandlerFunc) {
+	s.customMiddlewares = customMiddlewares
```
rbeuque74, line 61: I wonder if multiple plugins starts to add middlewares, they will break previous plugins middlewares.

```suggestion
	s.customMiddlewares = append(s.customerMiddlewares, customMiddlewares...)
```
This suggestion might fix it. What do you think @zhouzhuojie @loopfz ?

api/server.go
```diff
@@ -54,6 +55,12 @@ func (s *Server) WithAuth(authProvider func(*http.Request) (string, error)) {
 	}
 }
 
+// WithCustomMiddlewares sets an array of customized gin middlewares.
+// It helps for init plugins to include these customized middlewares in the api server
+func (s *Server) WithCustomMiddlewares(customMiddlewares ...gin.HandlerFunc) {
+	s.customMiddlewares = append(s.customerMiddlewares, customMiddlewares...)
```
rbeuque74, line 61: Typo, my bad
```suggestion
	s.customMiddlewares = append(s.customMiddlewares, customMiddlewares...)
```

zhouzhuojie: good catch

loopfz:
LGTM, thanks for the contrib !
Merging now

PR #178, opened 2020-09-09
feat: add insecure_skip_verify option to the builtin HTTP plugin

wI2L:


pkg/plugins/builtin/http/http.go
```diff
@@ -193,10 +195,23 @@ func exec(stepName string, config interface{}, ctx interface{}) (interface{}, in
...
+	if cfg.InsecureSkipVerify != "" {
+		insecureSkipVerify, err = strconv.ParseBool(cfg.InsecureSkipVerify)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to parse insecure_skip_verify: %s", err)
+		}
+	}
+	httpClient := httputil.NewHTTPClient(httputil.HTTPClientConfig{
+		Timeout:        td,
+		FollowRedirect: fr,
+		Transport: &http.Transport{
```
rbeuque74: We should clone http.DefaultTransport to keep all good timeout values

pkg/plugins/builtin/http/http.go
```diff
@@ -193,10 +195,23 @@ func exec(stepName string, config interface{}, ctx interface{}) (interface{}, in
...
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: insecureSkipVerify},
```
rbeuque74: Not sure about what is a good DefaultTLSClientConfig to keep HTTP2 enabled here.

rbeuque74 (changes requested):

rbeuque74 (approved):

rbeuque74:
Let's maybe bump the minor version as we are breaking compatibility for people building with Go 1.12 (there should be none)

PR #180, opened 2020-09-21
fix: unable to edit task that have read-only tags

rbeuque74:
Bug fix

When a subtask is spawned, some default tags are assigned to the subtask, and they are only read-only.
We can't edit the task anymore without removing the read-only tag, which is annoying.

Handle correctly the task edition without touching the read-only tags.

No

* **Other information**:

### DIFF
--- a/api/handler/task.go
+++ b/api/handler/task.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ovh/utask/models/task"
 	"github.com/ovh/utask/models/tasktemplate"
 	"github.com/ovh/utask/pkg/auth"
+	"github.com/ovh/utask/pkg/constants"
 	"github.com/ovh/utask/pkg/taskutils"
 	"github.com/ovh/utask/pkg/utils"
 )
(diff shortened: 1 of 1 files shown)


api/handler/task.go
```diff
@@ -270,9 +271,13 @@ func UpdateTask(c *gin.Context, in *updateTaskIn) (*task.Task, error) {
 	t.SetInput(clearInput)
 	t.SetWatcherUsernames(in.WatcherUsernames)
 
-	if err := utils.ValidateTags(in.Tags); err != nil {
-		return nil, err
+	// validate read-only tags
+	v, readOnlyTagUpdated := in.Tags[constants.SubtaskTagParentTaskID]
+	oldValue, readOnlyTagInTask := t.Tags[constants.SubtaskTagParentTaskID]
+	if (readOnlyTagUpdated && (!readOnlyTagInTask || oldValue != v)) || (!readOnlyTagUpdated && readOnlyTagInTask) {
```
wI2L, line 277: Not sure about the second part of the condition:`(!readOnlyTagUpdated && readOnlyTagInTask)`. This implies that during an update, the read-only tag must be given and be the same value. Is it intentional ?

loopfz: It seems right to me, their absence would mean a deletion otherwise?

wI2L:
Yes, you're right, I was thinking of a PATCH behavior.

Le lun. 21 sept. 2020 à 20:04, Thomas Schaffer <notifications@github.com> a
écrit :

api/handler/task.go
rbeuque74: Exactly :+1:

rbeuque74:
Can I consider that no comments means LGTM ? :D

wI2L (approved):

PR #181, opened 2020-09-21
subtask: change step state to TO_RETRY instead of SERVER_ERROR while not complete

rbeuque74:
When subtask is not complete, the 'subtask' step that created the
subtask falls into SERVER_ERROR, state that can be confusing for
end-users.
errors.NotProvisioned class allows plugins to return state 'TO_RETRY',
which will be handled correctly by the remediation loop.


loopfz (approved):
Lgtm

PR #182, opened 2020-09-21
fix: task: always return input in GET /task

rbeuque74:
When a task have no inputs, we should return the input property, to
provide a stable API for the dashboard UI

Key is missing

Key is present now

No

* **Other information**:


wI2L:
IMO, this should be handled by the API. The Javascript would do the same thing for `undefined`, `null`, or an empty array `[]`.

loopfz (approved):

PR #190, opened 2020-10-09
feat: Make a step loop while a condition is true

Aksamyt:

A new type of condition: `loop`. This forces a step to be evaluated again while the condition is true. There is also a new runnable state, `IN_LOOP`.

Here is a trivial example where a step counts to a number:
```yaml
steps:
    count:
        description: counts
        conditions:
        - type: loop
          if:
          - value: '{{ .step.this.output }}'
            operator: LT
            expected: '{{ .input.max }}'
        action:
            type: echo
            configuration:
                output: '{{ add (default .step.count.output 0) 1 }}'
```

Currently there is no easy way to write a loop with a condition.
 1. `foreach` is not suitable for cases where we don’t know how many items we need to iterate on.
2. The closest way we can do this is to have a step setting its own state to `TO_RETRY` with a `check` condition. However, the backoff duration quickly adds up and the result is less than ideal.
3. Another solution would be to write a plugin or a script, though we won’t benefit from important μTask features like retrying and blocking.

A new condition type: `loop`.

No

* **Other information**:

Nothing of importance, but while testing this feature I came across an interesting behaviour: if a step has a `foreach` property and a `loop` condition, then the children will loop independently from eachother. This might be used to factorise similar steps together.


PHD59fr (dismissed):

engine/step/step.go
```diff
@@ -32,6 +32,7 @@ const (
 const (
 	StateAny           = "ANY" // wildcard
 	StateTODO          = "TODO"
+	StateInLoop        = "IN_LOOP"
```
wI2L, line 35: Personal preference, but I'd rename it to `StateLooping`, with value `LOOPING` for consistency with existing states.

engine/step/step.go
```diff
@@ -72,6 +73,7 @@ var (
 // A step can be configured to evaluate "conditions" before and after the action is performed:
 // - a "skip" condition will be run before and might determine that the step's action can be skipped entirely
 // - a "check" condition will be run after the action, and can control execution flow by examining
+// - a "loop" condition will cause the step to run again
```
wI2L, line 76: We should probably mention the condition on which the step is ran again, meaning, when the AND'ed conditions equals to `true`.

engine/step/step.go
```diff
@@ -493,11 +495,16 @@ func AfterRun(st *Step, values *values.Values, ss StateSetter) {
 				break
 			}
 		}
-		for step, state := range sc.Then {
-			if step == stepRefThis {
-				step = st.Name
+		switch sc.Type {
+		case condition.LOOP:
+			ss(st.Name, StateInLoop, sc.Message)
```
wI2L, line 500: `sc.Message` seems inappropriate for me in this case. During the execution, the state of the step could contains the current loop iteration number, maybe?

wI2L (changes requested):
I'm in favor of this feature, as it nicely covers many usecase, like iterating cursor-paginated APIs.
However, I think we need to implement configurable parameters for the following:
- Maximum number of iterations allowed. We could use a sane default in the code, and allow users to explicitely remove that limit if they set the value of tha parameter to `0` or `-1`.
IMO, a step that has a `loop` condition shouldn't be retried if one of the action runs fails, so the `max_retries` property of the step could be used to control the maximum number of iterations.
- Set a delay between each execution of the step.

Also, a the end of a loop's execution, it could be handy to have some kind of metadata, like the number of runs that were executed.

engine/step/step.go
Aksamyt: Do you mean something like this?
```go
if st.MaxRetries < 0 || st.TryCount <= st.MaxRetries {
	state := fmt.Sprintf("%s#%d", StateLooping, st.TryCount)
	ss(st.Name, state, "")
} else {
	ss(st.Name, StateDone, "")
	logrus.Debugf("AfterRun: Step [%s] loop condition: max_retries (%d) exceeded", st.Name, st.MaxRetries)
}
```

engine/morejson/positive_duration.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"time"
+)
+
+// PositiveDuration is a wrapper around time.Duration.
+// It restricts its values to positive ones only, and it implements the
+// json.Marshaler and json.Unmarshaler interfaces.
+type PositiveDuration struct {
```
wI2L, line 12: I don't think we need a cusom time. You coul either check the value of the `time.Duration` value at runtime, or simply ignore it, since it is only used by `time.Sleeps`, which state in its doc:
> A negative or zero duration causes Sleep to return immediately.

Aksamyt: Right, didn’t notice that. What should I do?
- scrap this type, set `step.LoopDelay`’s type to `string`, and parse the duration in `AfterRun` (with a warning log if the parsing fails)?
- keep this type but refactor it?

rbeuque74: What about: `step.LoopDelay` as a `time.Duration`, and validate inside the function `step.ValidAndNormalize` (https://github.com/ovh/utask/blob/master/engine/step/step.go#L512) that the value is not negative ?
That's mean that the error will be triggered at boot time, instead of run time; and get rides of this new type ?

engine/step/step.go
```diff
@@ -493,13 +501,30 @@ func AfterRun(st *Step, values *values.Values, ss StateSetter) {
...
+				ss(step, state, sc.Message)
 			}
-			ss(step, state, sc.Message)
 		}
 	}
+
+	// check the status after the conditions’ evaluation in case a "check"
+	// condition changed the status
+	if strings.HasPrefix(st.State, StateLooping) {
+		time.Sleep(st.LoopDelay.Duration)
```
rbeuque74, line 526: I'm not sure to understand why this might be needed

engine/step/step.go
```diff
@@ -493,13 +501,30 @@ func AfterRun(st *Step, values *values.Values, ss StateSetter) {
 				break
 			}
 		}
-		for step, state := range sc.Then {
-			if step == stepRefThis {
-				step = st.Name
+		switch sc.Type {
+		case condition.LOOP:
+			if st.MaxRetries < 0 || st.TryCount <= st.MaxRetries {
+				state := fmt.Sprintf("%s%d", StateLooping, st.TryCount)
```
rbeuque74, line 507: I wonder why we need to have the `TryCount` as the state name here.
Why don't go directly with `LOOPING`, and use the `TryCount` to know how many iterations as been performed ?

engine/step/step.go
rbeuque74: Ok, I saw @wI2L comment

> Set a delay between each execution of the step.

Do you have more insights about having a pause between execution ?
Having a `time.Sleep` in the main thread will prevent others steps collection, and could lead to data loss if the application receive a `SIGTERM`.

rbeuque74:
On a global note, I would like to have this feature in uTask, thanks for your contribution! :)

loopfz (changes requested):
Cool feature overall, thanks for the contrib!

Overall I find that **loop** conditions look very similar to **check** conditions: they both happen at the same time (after execution, only if in a final state).
The difference is that they have an implied effect (eg RETRY bypassing the mandatory backoff).

For clarity, I suggest implementing the loop logic as a different step_state instead of a condition type.

eg

```yaml
if:
  - value: '{{ .step.this.output }}'
    operator: LT
    expected: '{{ .input.max }}'
then:
  this: RETRY_NOW
```

For the delay, I suggest naming it **execution_delay** (type time.Duration, sanitize its value between acceptable boundaries e.g. min 0 / max XX (20?) secs; we dont want a step staying with a RUNNING state for too long because it makes it harder to recover from crashes) and doing a time.Sleep in step.Run, after the go func() (to get out of the main control goroutine) and before the plugin exec (and semaphore slot acquire).

loopfz (changes requested):

engine/step/step.go
```diff
@@ -497,7 +505,12 @@ func AfterRun(st *Step, values *values.Values, ss StateSetter) {
 			if step == stepRefThis {
 				step = st.Name
 			}
-			ss(step, state, sc.Message)
+			if state == StateRetryNow && st.MaxRetries >= 0 && st.TryCount > st.MaxRetries {
+				ss(st.Name, StateDone, "")
```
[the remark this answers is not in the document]
rbeuque74: Agree. I guess this is not mandatory, as it will be picked up in the next iteration by `step.Run`

wI2L (dismissed):

ovh-cds:
CDS Report test#861.0 ✘  
* Stage 1
  * Test ✘

wI2L (approved):

rbeuque74 (approved):

rbeuque74:
Thanks for your contribution @Aksamyt, sorry for the delay !

Aksamyt:
No problem! Hope it will be useful to many people

PR #192, opened 2020-11-06
feat: Update prometheus metrics and add duration metrics

amstuta:
Signed-off-by: Arthur Amstutz <18141571+amstuta@users.noreply.github.com>

  - Simplify registration of previously existing statistics to Prometheus
  - Add new statistics: duration between task creation and resolution creation, execution time, and complete time taken between task creation and task completion

The number of tasks by state is register state by state in Prometheus (one Gauge for each state).

A GaugeVec is used to store all the Gauges for the number of tasks by state.

  - Database schema change: the column "created" is added in table "resolution" to register the timestamp when a resolution is created
  - Users that retrieved the number of tasks by state in Prometheus will have to change the way of getting each individual stat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ We will list in this document any breaking changes between versions, that requir
 
 ## Breaking changes
 
-### v1.9.0
+### v1.10.0
 #### SQL
 - `005_resolution_creation_timestamp.sql` migration file should be applied while upgrading. It adds a column `created` in the `resolution` table to keep track of the resolution creation timestamp.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ We will list in this document any breaking changes between versions, that requir
 
 ## Breaking changes
 
+### v1.9.0
+#### SQL
+- `005_resolution_creation_timestamp.sql` migration file should be applied while upgrading. It adds a column `created` in the `resolution` table to keep track of the resolution creation timestamp.
+
+#### Prometheus metrics
+- The Prometheus gauges used to record the number of tasks by state haven been moved to a single `GaugeVec` named `utask_task_state`.
+
 ### v1.5.0 (2020-04-09)
 ##### Templating
 - `jsonmarshal` function has been __deleted__. It was obsolete since the `sprig` package used by utask provides many JSON related functions (e.g. `toJson`).
--- a/models/task/stats.go
+++ b/models/task/stats.go
@@ -1,15 +1,48 @@
 package task
 
 import (
+	"time"
+
 	"github.com/juju/errors"
 	"github.com/loopfz/gadgeto/zesty"
 	"github.com/ovh/utask/db/pgjuju"
 	"github.com/ovh/utask/db/sqlgenerator"
+	"github.com/ovh/utask/pkg/now"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	validationTimes = promauto.NewSummaryVec(prometheus.SummaryOpts{Name: "utask_valid_times"}, []string{"template"})
+	completionTimes = promauto.NewSummaryVec(prometheus.SummaryOpts{Name: "utask_complete_times"}, []string{"template"})
+	executionTimes  = promauto.NewSummaryVec(prometheus.SummaryOpts{Name: "utask_exec_times"}, []string{"template"})
 )
 
 type stateCount struct {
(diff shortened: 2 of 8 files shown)


loopfz (approved):
lgtm; need to plan a changelog update separately (breaking changes = sql schema + prom series previous label change)
@rbeuque74 lgty?

models/task/stats.go
```diff
@@ -1,17 +1,50 @@
...
+	validationTimes.WithLabelValues(templateName).Observe(duration)
+}
+
+// RegisterTaskTime computes the execution duration and the complete duration
+// (from creation to completion) of a task. These metrics are then pushed to Prometheus.
+func RegisterTaskTime(templateName string, tackCreation, resCreation time.Time) {
+	currentTime := now.Get()
+
+	// Time taken since task creation
+	completeTime := currentTime.Sub(tackCreation).Seconds()
```
rbeuque74, line 40:
```suggestion
func RegisterTaskTime(templateName string, taskCreation, resCreation time.Time) {
	currentTime := now.Get()

	// Time taken since task creation
	completeTime := currentTime.Sub(taskCreation).Seconds()
```

rbeuque74 (changes requested):

sql/migrations/005_resolution_creation_timestamp.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+
+-- Adds the resolution creation timestamp on resolution table
+ALTER TABLE "resolution" ADD COLUMN "created" TIMESTAMP with time zone DEFAULT now() NOT NULL;
```
rbeuque74, line 4: Wonder if the migration process should not manage differently already created/already finished task.
If we have 200 tasks in the database, with state DONE, at the time we run the `created` value will have a value after `last_start` and `last_stop`.
Should we put the value to `last_start` for existing task?

Making the query to something like
```sql
ALTER TABLE "resolution" ADD COLUMN "created" TIMESTAMP with time zone;
UPDATE "resolution" SET created = last_start;
ALTER TABLE "resolution" ALTER COLUMN "created" SET NOT NULL;
```

@loopfz @wI2L thoughts?
```

**Comment by rbeuque74** [comment]
> lgtm; need to plan a changelog update separately (breaking changes = sql schema + prom series previous label change)
> @rbeuque74 lgty?

OK to add a section into https://github.com/ovh/utask/blob/master/CHANGELOG.md.
@amstuta you can check the section SQL of v1.4.0 for example about SQL migration.
We will release this under v1.9.0

I will apply the typo change to the pull-request also.

File: sql/migrations/005_resolution_creation_timestamp.sql (hunk 2 above)
Comment (rbeuque74) [reply sql/migrations/005_resolution_creation_timestamp.sql:4]: @loopfz @wI2L  up

Comment (loopfz) [reply sql/migrations/005_resolution_creation_timestamp.sql:4]: last_stop sounds better? but yeah, agree

Comment (amstuta) [reply sql/migrations/005_resolution_creation_timestamp.sql:4]: I agree. Since the `created` field represents the resolution creation, shouldn't `last_start` be closer to it ?

File: CHANGELOG.md [hunk 3]
```diff
@@ -6,6 +6,13 @@ We will list in this document any breaking changes between versions, that requir
 
 ## Breaking changes
 
+### v1.9.0
```
rbeuque74, line 9:
```suggestion
### v1.10.0
```

rbeuque74 (approved):

sql/migrations/005_resolution_creation_timestamp.sql
```diff
@@ -0,0 +1,10 @@
+-- +migrate Up
+
+-- Adds the resolution creation timestamp on resolution table and sets its value to last_start
+ALTER TABLE "resolution" ADD COLUMN "created" TIMESTAMP with time zone;
+UPDATE "resolution" SET created = last_start;
+ALTER TABLE "resolution" ALTER COLUMN "created" SET NOT NULL;
```
rbeuque74, line 6: I had a new thought about the migration procedure: if we don't have the `DEFAULT now();` here, then once the SQL migration is run, all the µTask must run on the new version, otherwise any `INSERT` into the `resolution` table will fail, due to the `NOT NULL` constraint.

Also, that's the implementation inside the `sql/schema.sql`.

```suggestion
ALTER TABLE "resolution" ALTER COLUMN "created" SET NOT NULL, ALTER COLUMN "created" SET DEFAULT now();
```

rbeuque74 (changes requested):

rbeuque74 (approved):

PR #196, opened 2020-11-15
resource limiting: complete rework to handle context cancelation and native rate-limiting on a template

rbeuque74:
This introduces a complete rework of resource usage by implementing
contexts, so the lock can be cancelled, in case the lock takes too long
to be obtained, or if the instance is shutting-down.

It also introduces a native resource on all resolution based on the
template name: it allows to reduce the concurrent execution of a given
template, or to completely stop the execution of a template if it behave
incorrectly.

This also fix some dead-lock that can be triggered, if a resource limit is set to 0,
as the lock will never be obtained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,7 @@ Note that the operators `IN` and `NOTIN` expect a list of acceptable values in t
 - `dependencies`: a list of step names on which this step waits before running
 - `custom_states`: a list of personnalised allowed state for this step (can be assigned to the state's step using `conditions`)
 - `retry_pattern`: (`seconds`, `minutes`, `hours`) define on what temporal order of magnitude the re-runs of this step should be spread (default = `seconds`)
+- `resources`: a list of resources that will be used during the step execution, to control and limit the concurrent execution of the step (more information in [the resources section](#resources)).
 
 <p align="center">
 <img src="./assets/img/utask_backoff.png" width="70%">
@@ -480,14 +481,16 @@ The `pre_hook` field of a step can be set to define an action that is executed b
 doSomeAuthPost:
   pre_hook:
     type: http
-    method: "GET"
-    url: "https://myAwesomeApi/otp"
+    configuration:
+      method: "GET"
+      url: "https://example.org/otp"
   action:
     type: http
-    method: "POST"
-    url: "https://myAwesomeApi/doSomePost"
-    headers:
-      X-Otp: "{{ .pre_hook.output }}"
+    configuration:
+      method: "POST"
+      url: "https://example.org/doSomePost"
+      headers:
+        X-Otp: "{{ .pre_hook.output }}"
 ```
 
 #### Functions <a name="functions"></a>
@@ -586,6 +589,39 @@ This output can be then passed to another step in json format:
 ```yaml
 foreach: '{{.step.prefixStrings.children | toJson}}'
 ```
+
+#### Resources <a name="resources"></a>
+
+Resources are a way to restrict the concurrency factor of certain operations, to control the throughput and avoid dangerous behavior e.g. flooding the targets.
+
+High level view:
+
+- For each action to execute, a list of target `resources` is determined. (see later)
+- In the µTask configuration, numerical limits can be set to each _resource_ label. This acts as a semaphore, allowing a certain number of concurrent slots for the given _resource_ label. If no limit is set for a resource label, the previously mentionned target resources have no effect. Limits are declared in the `resource_limits` property.
+
+The target _resources_ for a step can be defined in its YAML definition, using the `resources` property.
+
+```yaml
+steps:
+  foobar:
+    description: A dummy step, that should not execute in parallel
+    resources: ["myLimitedResource"]
+    action:
+      type: echo
+      configuration:
+        output:
+          foobar: fuzz
+```
+
+Alternatively, some target resources are determined automatically by µTask Engine:
+
+- When a task is run, the resource `template:my-template-name` is used automatically.
+- When a step is run, the plugin in charge of the execution automatically generates a list of resources. This includes generic resources such as `socket`, `url:www.example.org`, `fork`...
+ allowing the µTask administrator to set-up generic limits such as `"socket": 48` or `"url:www.example.org": 1`.
+
+Each builtin plugins declares resources which can be discovered using the _README_ of the plugin (example for [_http_ plugin](./pkg/plugins/builtin/script/README.md#Resources)).
+
+Declared `resource_limits` must be positive integers. When a step is executed, if the number of concurrent executions is reached, the µTask Engine will wait for a slot to be released. If the resource is limited to the `0` value, then the step will not be executed and is set to `TO_RETRY` state, it will be run once the instance allows the execution of its resources. The default time that µTask Engine will wait for a resource to become available is `1 minute`, but it can be configured using the `resource_acquire_timeout` property.
 
 ### Task templates validation
 
--- a/api/handler/resolution.go
+++ b/api/handler/resolution.go
@@ -281,7 +281,24 @@ func RunResolution(c *gin.Context, in *runResolutionIn) error {
 
 	logrus.WithFields(logrus.Fields{"resolution_id": r.PublicID}).Debugf("Handler RunResolution: manual resolve %s", r.PublicID)
 
-	return engine.GetEngine().Resolve(in.PublicID, nil)
+	ch := make(chan struct{})
+	go func() {
+		err = engine.GetEngine().Resolve(in.PublicID, nil)
+		close(ch)
+	}()
+
+	timeout := time.NewTicker(5 * time.Second)
+	defer timeout.Stop()
+
+	// manual resolution can be blocked by a lock acquisition on the Execution pool
+	// waiting for 5 seconds to get a response, otherwise let's consider the task will
+	// start correctly when the Execution pool gets available, and prevent API thread to be blocked
+	select {
+	case <-ch:
+		return err
+	case <-timeout.C:
+		return nil
+	}
 }
 
 type extendResolutionIn struct {
--- a/api/server.go
+++ b/api/server.go
@@ -106,7 +106,7 @@ func (s *Server) ListenAndServe() error {
 		logrus.Info("Shutting down...")
 		cancel()
 
-		if err := srv.Shutdown(ctx); err != nil {
+		if err := srv.Shutdown(context.Background()); err != nil {
 			logrus.Fatal(err)
 		}
 	}()
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -36,8 +36,8 @@ var (
 	eng Engine
 
 	// Used for stopping the current Engine
-	stopRunningSteps chan struct{}
-	gracePeriodEnd   chan struct{}
+	shutdownCtx    context.Context
+	gracePeriodEnd chan struct{}
 )
 
 // Engine is the heart of utask: it is the active process
@@ -97,13 +97,11 @@ func Init(ctx context.Context, wg *sync.WaitGroup, store *configstore.Store) err
 	}
 
 	// channels for handling graceful shutdown
-	stopRunningSteps = make(chan struct{})
+	shutdownCtx = ctx
 	gracePeriodEnd = make(chan struct{})
 	eng.wg = wg
 	go func() {
-		<-ctx.Done()
-		// Stop running new steps
-		close(stopRunningSteps)
+		<-shutdownCtx.Done()
 
 		// Wait for the grace period to end
 		time.Sleep(3 * time.Second)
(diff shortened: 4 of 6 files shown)


README.md
```diff
@@ -563,6 +566,34 @@ This output can be then passed to another step in json format:
 foreach: '{{.step.prefixStrings.children | toJson}}'
 ```
 
+#### Resources <a name="resources"></a>
+
+Resources is a way to limit the number of current execution of tasks or steps within an µTask instance. When a lot of tasks are launched at the same time, a high throughput from µTask can hurt some components of your infrastructure, making them unavailable.
```
Alkorin, line 571:
```suggestion
Resources is a way to limit the number of concurrent execution of tasks or steps within an µTask instance. When a lot of tasks are launched at the same time, a high throughput from µTask can hurt some components of your infrastructure, making them unavailable.
```

README.md
```diff
@@ -563,6 +566,34 @@ This output can be then passed to another step in json format:
...
+
+Defining `resources` on a µTask instance, on a template, or on a step is a way to limit the number of concurrent executions on µTask side, to make sure that the targetted component stays healthy.
```
Alkorin, line 573:
```suggestion
Defining `resources` on a µTask instance, on a template, or a step is a way to limit the number of concurrent executions on µTask side, to make sure that the targetted component stays healthy.
```

README.md
```diff
@@ -563,6 +566,34 @@ This output can be then passed to another step in json format:
...
+    description: A dummy step, that should not execute in parallel
+    resources: [limitedResourceFoobar]
+    action:
+      type: echo
+      configuration:
+        output:
+          foobar: fuzz
+```
+
+Every time the step `foobar` will be executed, µTask engine will verify that the maximum concurrent limit is not exceeded, and acquire one slot. The slot will be hold until the step finishes its execution. If all slots for a resource are hold, then the current execution will wait until one slot is released, or if timeout is reached (defined in `resource_acquire_timeout` configuration, or default `1 minute`). If the resource is not declared in the `resource_limits` configuration, then, no limitation will be applied. If the limitation value is set to `0`, then, the step will not wait, and will be set to state `TO_RETRY` to be schedule later.
```
Alkorin, line 591:
```suggestion
Every time the step `foobar` will be executed, µTask engine will verify that the maximum concurrent limit is not exceeded, and acquire one slot. This slot will be held until the step finishes its execution. If all slots for a resource are held, then the current execution will wait until one slot is released, or if the timeout is reached (defined in `resource_acquire_timeout` configuration, or default `1 minute`). If the resource is not declared in the `resource_limits` configuration, then, no limitation will be applied. If the limit value is set to `0`, then, the step will not wait and will be set to state `TO_RETRY` to be scheduled later.
```

README.md
```diff
@@ -563,6 +566,34 @@ This output can be then passed to another step in json format:
...
+
+Templates declare a builtin resource, named `template:my-template-name` (with `my-template-name` being the template name): this allows to limit the number of concurrent execution of tasks from the given template. In that way, a way that consumes a lot of CPU/RAM/IO can be restrain to save the system resources, or to disable completely the execution of tasks from a given template.
```
Alkorin, line 593:
```suggestion
Templates declare a builtin resource, named `template:my-template-name` (with `my-template-name` being the template name): this allows to limit the number of concurrent execution of tasks from the given template. In that way, a way that consumes a lot of CPU/RAM/IO can be restrained to save the system resources or to disable completely the execution of tasks from a given template.
```

rbeuque74:
Thank you for the README review, it's appreciated :)

api/handler/resolution.go
```diff
@@ -252,7 +252,23 @@ func RunResolution(c *gin.Context, in *runResolutionIn) error {
...
+
+	// manual resolution can be blocked by a lock acquisition on the Execution pool
+	// waiting for 5 seconds to get a response, otherwise let's consider the task will
+	// start correctly when the Execution pool gets available, and prevent API thread to be blocked
+	select {
+	case <-ch:
+		return err
+	case <-chTimeout:
+		return nil
+	}
```
Alkorin, line 271: time.After will leak the timer until the timer fires. If there is a lot of calls and it is unlikely that the timer fires, it is better to explicitly stop it to reduce resource consumption

```suggestion
	timeout := time.NewTimer(5 * time.Second)
	defer timeout.Stop()

	// manual resolution can be blocked by a lock acquisition on the Execution pool
	// waiting for 5 seconds to get a response, otherwise let's consider the task will
	// start correctly when the Execution pool gets available, and prevent API thread to be blocked
	select {
	case <-ch:
		return err
	case <-timeout.C:
		return nil
	}
```

engine/engine.go
```diff
@@ -211,14 +209,40 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
 
 	res.Values.SetConfig(e.config)
 
-	// all ready, run remaining steps
-
-	utask.AcquireExecutionSlot()
-
+	// checking if all resources are available before starting the resolution
+	// first, checking if we have a custom semaphore, for example, crashed instance recover
+	// putting this semaphore first, cause the crashed instance limitation is always lower than the execution pool
```
Alkorin, line 214:
```suggestion
	// check if all resources are available before starting the resolution
	// first, check if we have a custom semaphore, for example, crashed instance recover
	// putting this semaphore first, cause the crashed instance limitation is always lower than the execution pool
```

engine/engine.go
```diff
@@ -211,14 +209,40 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
...
 	if sm != nil {
-		sm.Acquire(context.Background(), 1)
+		if err := sm.Acquire(shutdownCtx, 1); err != nil {
+			debugLogger.Debugf("Engine: launchResolution() %s acquire resource: instance is shutting down", res.PublicID)
+			return nil, errors.New("instance is shutting down")
+		}
+	}
+	// second, checking if we have restricted resource on the given template
+	// template could be completely desactivated, as a "dead resource", if it's the case, we need to exit because
+	// the limit won't change until instance rebooted
```
Alkorin, line 223:
```suggestion
	// second, check if we have a restricted resource on the given template
	// template could be completely deactivated as a "dead resource". If it's the case, we need to exit because
	// the limit won't change until next instance's reboot
```

utask.go
```diff
@@ -172,24 +183,77 @@ func (c *Cfg) buildLimits() {
...
+	}
+
 	s := global.resourceSemaphores[name]
 	if s == nil {
-		return
+		return nil
+	}
+
+	semaphoreCtx := ctx
+	if global.resourceAcquireTimeoutDuration == 0 {
```
Alkorin, line 208: != 0 no ?

engine/engine.go
```diff
@@ -211,14 +209,40 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
...
+	if acquiredErr := utask.AcquireResource(shutdownCtx, "template:"+t.TemplateName); acquiredErr != nil {
+		if shutdownCtx.Err() != nil {
+			debugLogger.Debugf("Engine: launchResolution() %s acquire resource: instance is shutting down", res.PublicID)
+			return nil, errors.New("instance is shutting down")
+		}
+		debugLogger.Debugf("Engine: launchResolution() %s acquire resource %q: failed to acquire resource", res.PublicID, "template:"+t.TemplateName)
```
Alkorin, line 229: Log acquiredErr ?

README.md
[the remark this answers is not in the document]
loopfz: Something like.

Resources are a way to restrict the concurrency factor of certain operations, to control the throughput and avoid dangerous behavior e.g. flooding the targets.

High level view:
- For each action to execute, a list of target `resources` is determined. (see later)
- In the uTask configuration, limits can be set for each `resource` label. This acts as a semaphore, allowing a certain number of concurrent slots for the given `resource` label. If no limit is set for a `resource` label, the previously mentionned target `resources` have no effect.

The target `resources` for a step can be defined in its YAML definition, using the `resources` property.
Alternatively, some target `resources` are determined automatically:
- When a task is run, the `resource` `template:my-template-name` is used automatically
- When a step is run, the plugin in charge of the execution automatically generates a list of `resources`. This includes generic `resources` such as `socket`, `url:www.google.com`, `fork`, ... allowing the uTask administrator to set-up generic limits such as `"socket": 48` or `"url:www.google.com": 1`.

engine/engine.go
```diff
@@ -211,14 +209,43 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
...
+		if err := commit(dbp, res, t); err != nil {
+			debugLogger.Debugf("Engine: launchResolution() %s acquire resource, FAILED TO COMMIT RESOLUTION: %s", res.PublicID, err)
+		}
+		if sm != nil {
+			sm.Release(1)
+		}
+		return nil, fmt.Errorf("can't acquire lock for template %q: %s", t.TemplateName, err)
+	}
+	// finally, acquire the execution slot
+	if acquiredErr := utask.AcquireExecutionSlot(shutdownCtx); acquiredErr != nil {
```
loopfz, line 243: now that we represent all concepts via resource labels e.g. template, socket, fork, url:XXX... wouldnt it make sense to have a built-in "task" or something to avoid considering these "execution" slots in a special way ?

engine/engine.go
```diff
@@ -211,14 +209,43 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
...
-		sm.Acquire(context.Background(), 1)
+		if err := sm.Acquire(shutdownCtx, 1); err != nil {
+			debugLogger.Debugf("Engine: launchResolution() %s acquire resource: instance is shutting down", res.PublicID)
+			return nil, errors.New("instance is shutting down")
+		}
+	}
+	// second, check if we have a resource limit on the current template
+	// template could be completely deactivated as a "dead resource". If it's the case, we need to exit because
+	// the limit won't change until next instance's reboot.
+	if acquiredErr := utask.AcquireResource(shutdownCtx, "template:"+t.TemplateName); acquiredErr != nil {
```
loopfz, line 224: You dont specify why you'd want to pass the shutdown context to the semaphore acquire calls?

From the doc: ```blocking until resources are available or ctx is done. ... If ctx is already done, Acquire may still succeed without blocking. ```

Wouldnt that possibly trigger weird storms in the seconds before shutdown, with all semaphores suddenly becoming non-blocking ?

utask.go
```diff
@@ -157,39 +163,99 @@ type DatabaseConfig struct {
 
 func (c *Cfg) buildLimits() {
 	c.resourceSemaphores = make(map[string]*semaphore.Weighted)
+	c.deadResources = make(map[string]struct{})
 
 	for k, v := range c.ResourceLimits {
-		c.resourceSemaphores[k] = semaphore.NewWeighted(int64(v))
+		if v <= 0 {
```
loopfz, line 169: Would it make sense to have 0 mean "nothing allowed" and negative values mean "disregard that config, return to default behavior a.k.a not care about limit" ?

Right now if I want to disable a limit, I have to take it out of the cfg. This would allow me to flip back and forth.

loopfz (changes requested):

engine/engine.go
rbeuque74: I'm giving the shutdownCtx to the semaphore acquire calls to release the "sleep" process, and then let the task be released.

> Wouldnt that possibly trigger weird storms in the seconds before shutdown, with all semaphores suddenly becoming non-blocking ?

It's possible that the semaphore succeed yes, but it won't go that far, as we check the shutdownCtx in the `runAvailableSteps` functions, meaning that no step will be evaluated. We will just give the task a good closing, otherwise, we would keep all those tasks in a `RUNNING` state, meanwhile it were only locked against the semaphore.

utask.go
rbeuque74: We could have that.
I just didn't wanted to change the current behaviour too much, as the pull request is already big.
We allow the negative value right now, but that creates semaphores that are blocking infinitely, until the context is cancelled (current context being `context.Background()`) ... https://github.com/golang/sync/blob/master/semaphore/semaphore.go#L49

engine/engine.go
rbeuque74: We could have a reserved keyword in the future for this execution slot.

README.md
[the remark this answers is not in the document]
rbeuque74: done

README.md
```diff
@@ -587,6 +590,39 @@ This output can be then passed to another step in json format:
 foreach: '{{.step.prefixStrings.children | toJson}}'
 ```
 
+#### Resources <a name="resources"></a>
```
wI2L, line 593: What is the link used for ?

api/handler/resolution.go
```diff
@@ -281,7 +281,24 @@ func RunResolution(c *gin.Context, in *runResolutionIn) error {
 
 	logrus.WithFields(logrus.Fields{"resolution_id": r.PublicID}).Debugf("Handler RunResolution: manual resolve %s", r.PublicID)
 
-	return engine.GetEngine().Resolve(in.PublicID, nil)
+	ch := make(chan struct{})
+	go func() {
+		err = engine.GetEngine().Resolve(in.PublicID, nil)
+		ch <- struct{}{}
```
wI2L, line 287: just close it.

api/server.go
```diff
@@ -106,7 +106,7 @@ func (s *Server) ListenAndServe() error {
 		logrus.Info("Shutting down...")
 		cancel()
 
-		if err := srv.Shutdown(ctx); err != nil {
+		if err := srv.Shutdown(context.Background()); err != nil {
```
wI2L, line 109: No timeout configured on a new context ?

engine/step/step.go
```diff
@@ -384,8 +390,18 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, stepValues *values.Val
 		// Wait the prehook execution is done
 		preHookWg.Wait()
 
+		// after pre-hook execution, let's verify if we are not shutting down the instance
+		// in that case, we can just put the step in ToRetry instead of starting the main execution of the step
+		select {
+		case <-shutdownCtx.Done():
+			st.State = StateToRetry
+			go noopStep(st, stepChan)
```
wI2L, line 398: Shoudn't use a synchronous call of `noopStep` instead ? You're firing a goroutine while the instance is shutting down, that goroutine won't be accounted for and main process won't wait for it to finish I think.

utask.go
```diff
@@ -165,39 +171,99 @@ type DatabaseConfig struct {
...
+	if c.MaxConcurrentExecutions != nil {
+		return *c.MaxConcurrentExecutions
 	}
+	return defaultMaxConcurrentExecutions
 }
 
+var (
+	// ErrDeadResource is returned when a resource will never be available, as the max concurrent execution is set to 0, and there is no reason to wait
+	ErrDeadResource          = errors.New("resource is not available, as configured with 0 concurrent execution")
+	ErrFailedAcquireResource = errors.New("failed to acquire the requested resource")
```
wI2L, line 199: `ErrFailedAcquireResource` needs to be documented as well.

wI2L (changes requested):

README.md
rbeuque74: It's used line 359 to give a quick link to this section.

api/server.go
rbeuque74: It would be tricky to have a timed shutdown after that all our backgrounds goroutines finished to stop.
I prefer that we don't Fatal because Shutdown didn't stop in a given time.

PR #199, opened 2020-11-20
http: add mTLS for client authentication

rbeuque74:
Feature

Add configuration field to enable Mutual TLS client authentication.

### DIFF
--- a/pkg/plugins/builtin/http/README.md
+++ b/pkg/plugins/builtin/http/README.md
@@ -13,7 +13,7 @@ This plugin permorms an HTTP request.
 | `body`                 | a string representing the payload to be sent with the request                                                                                                                                                                                                    |
 | `headers`              | a list of headers, represented as (`name`, `value`) pairs                                                                                                                                                                                                        |
 | `timeout`              | timeout expressed as a duration (e.g. `30s`)                                                                                                                                                                                                                     |
-| `auth`                 | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or `bearer` field to enable Bearer Token Authorization                                                                                          |
+| `auth`                 | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or a `bearer` field to enable Bearer Token Authorization, or a `mutual_tls` object to enable Mutual TLS authentication                          |
 | `follow_redirect`      | if `true` (string) the plugin will follow up to 10 redirects (302, ...)                                                                                                                                                                                          |
 | `query_parameters`     | a list of query parameters, represented as (`name`, `value`) pairs; these will appended the query parameters present in the `url` field; parameters can be repeated (in either `url` or `query_parameters`) which will produce e.g. `?param=value1&param=value2` |
 | `trim_prefix`          | prefix in the response that must be removed before unmarshalling (optional)                                                                                                                                                                                      |
@@ -39,8 +39,17 @@ action:
         user: {{.config.basicAuth.user}}
         password: {{.config.basicAuth.password}}
       bearer: {{.config.auth.token}}
+      mutual_tls:
+        # a chain of certificates to identify the caller, first certificate in the chain is considered as the leaf, followed by intermediates
+        client_cert: {{.config.mtls.clientCert}}
+        # private key corresponding to the certificate
+        client_key: {{.config.mtls.clientKey}}
     # optional, string as boolean
     follow_redirect: "true"
+    # optional, defines additional root CAs to perform the call. can contains multiple CAs concatained together
+    root_ca: {{.config.mtls.rootca}}
+    # optional, string as boolean. indicates if server certificate must be validated or not.
+    insecure_skip_verify: "false"
     # optional, array of name and value fields
     query_parameters:
     - name: foo
--- a/pkg/plugins/builtin/http/http.go
+++ b/pkg/plugins/builtin/http/http.go
@@ -17,7 +17,6 @@ import (
 	"github.com/ovh/utask/pkg/plugins/builtin/httputil"
 	"github.com/ovh/utask/pkg/plugins/taskplugin"
 	"github.com/ovh/utask/pkg/utils"
-	"golang.org/x/net/http2"
 )
 
 // the HTTP plugin performs an HTTP call
@@ -27,15 +26,6 @@ var (
 		taskplugin.WithResources(resourceshttp),
 	)
 )
-
-var defaultUnsecureTransport http.RoundTripper
-
-func init() {
-	tr := http.DefaultTransport.(*http.Transport).Clone()
-	tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-	_ = http2.ConfigureTransport(tr)
-	defaultUnsecureTransport = tr
-}
 
 const (
 	// TimeoutDefault represents the default value that will be used for HTTP call, if not defined in configuration
@@ -56,6 +46,7 @@ type HTTPConfig struct {
 	QueryParameters    []parameter `json:"query_parameters,omitempty"`
 	TrimPrefix         string      `json:"trim_prefix,omitempty"`
 	InsecureSkipVerify string      `json:"insecure_skip_verify,omitempty"`
+	RootCA             string      `json:"root_ca,omitempty"`
 }
 
 // parameter represents either headers, query parameters, ...
@@ -66,14 +57,20 @@ type parameter struct {
 
 // auth represents HTTP authentication
 type auth struct {
-	Basic  authBasic `json:"basic"`
-	Bearer string    `json:"bearer"`
+	Basic     *authBasic `json:"basic"`
+	Bearer    *string    `json:"bearer"`
+	MutualTLS *mTLS      `json:"mutual_tls"`
 }
 
 // authBasic represents the embedded basic auth inside Auth struct
 type authBasic struct {
 	User     string `json:"user"`
 	Password string `json:"password"`
+}
+
+type mTLS struct {
+	ClientCert string `json:"client_cert"`
+	ClientKey  string `json:"client_key"`
 }
 
 func validConfig(config interface{}) error {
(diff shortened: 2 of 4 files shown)


pkg/plugins/builtin/http/README.md
```diff
@@ -39,6 +39,10 @@ action:
         user: {{.config.basicAuth.user}}
         password: {{.config.basicAuth.password}}
       bearer: {{.config.auth.token}}
+      mTLS:
+        root_ca: {{.config.mtls.rootca}}
```
wI2L, line 43: Even tho its pretty common to include a chain of certs in the same document, it may be worh mentioning it. It directly refers to the usage of `AppendCertsFromPEM` later.

pkg/plugins/builtin/http/http.go
```diff
@@ -68,6 +58,7 @@ type parameter struct {
 type auth struct {
 	Basic  authBasic `json:"basic"`
 	Bearer string    `json:"bearer"`
+	MTLS   mTLS      `json:"mTLS"`
```
wI2L, line 61: This applies to the new field, bu also the existing one. I suggest that we use pointers, to simplify the checks later in `validateConfig`. It would be easier to ensure that many auth methods are provided since they are mutually exclusive, without having to check every of their fields individually. Nothing prevent from keepin the current checks that ensure that all parameters required for an auth method in particular are present.

pkg/plugins/builtin/httputil/httputil.go
```diff
@@ -110,3 +113,42 @@ func defaultHTTPClientFactory(cfg HTTPClientConfig) HTTPClient {
...
+		t.TLSClientConfig.InsecureSkipVerify = v
+	}
+}
+
+func WithTLSClientAuth(cert tls.Certificate) func(*http.Transport) {
+	return func(t *http.Transport) {
+		if t.TLSClientConfig == nil {
+			t.TLSClientConfig = &tls.Config{}
+		}
+		t.TLSClientConfig.Certificates = []tls.Certificate{cert}
```
wI2L, line 141: Append instead to preserve existing certs ? Could that be necessary ? Guess not but the override tickled me, so just commenting for awarness.

pkg/plugins/builtin/httputil/httputil.go
```diff
@@ -110,3 +113,42 @@ func defaultHTTPClientFactory(cfg HTTPClientConfig) HTTPClient {
...
+	}
+}
+
+func WithTLSRootCA(caCert []byte) func(*http.Transport) {
+	return func(t *http.Transport) {
+		if t.TLSClientConfig == nil {
+			t.TLSClientConfig = &tls.Config{}
+		}
+		caCertPool := x509.NewCertPool()
```
wI2L, line 150: Same as for `WithTLSClientAuth`, if the cert poll already exist, maybe reuse it ?

pkg/plugins/builtin/http/http.go
wI2L, line 61: Hah, too bad Go struct fields naming rules got in your way 😄

wI2L (changes requested):

pkg/plugins/builtin/http/http.go
rbeuque74: Sad developer is sad.

pkg/plugins/builtin/http/README.md
rbeuque74: Regarding `root_ca`, I guess I shouldn't have put in here.
This is not linked to mTLS feature, this only comes to be able to provide another Root Certificate Authority while calling a distant service.
Maybe we can leave this configuration to the end-user, by adding the corresponding Root CA into the system trust-store, and let uTask relies on the system trust store.
What do you think ?

pkg/plugins/builtin/http/http.go
rbeuque74: Ok with that, I will change this.

pkg/plugins/builtin/httputil/httputil.go
rbeuque74: In this feature, this won't be needed, as only one client certificate can be provided.
As the `WithTLSClientAuth` could be used in the future, I will perform this change in case we need it.

pkg/plugins/builtin/httputil/httputil.go
rbeuque74: This method should not be re-used multiple time.
If someone want to alter the certPool with multiple certificates, I guess it should be done with one call to `WithTLSRootCA`, and providing multiples certificates as inputs.
Will clarify this in a comment.

pkg/plugins/builtin/http/README.md
wI2L: I think it can still be handy to specify a CA at runtime with this parameter. Not all usescases relies on the system trust-store.

pkg/plugins/builtin/httputil/httputil.go
wI2L: Yeah, I see where your're coming from on tat, but since the cert pool type allow to add certificates, does it really matter that we add to the existing pool instead of recreating one ?

pkg/plugins/builtin/http/http.go
loopfz: I think we have snake_case everywhere, `mTLS` looks out of place

pkg/plugins/builtin/http/README.md
```diff
@@ -13,7 +13,7 @@ This plugin permorms an HTTP request.
...
 | `headers`              | a list of headers, represented as (`name`, `value`) pairs                                                                                                                                                                                                        |
 | `timeout`              | timeout expressed as a duration (e.g. `30s`)                                                                                                                                                                                                                     |
-| `auth`                 | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or `bearer` field to enable Bearer Token Authorization                                                                                          |
+| `auth`                 | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or `bearer` field to enable Bearer Token Authorization, or `mTLS` field to enable Mutual TLS authentication                                     |
```
loopfz, line 16:
```suggestion
| `auth`                 | a single object composed of either a `basic` object with `user` and `password` fields to enable HTTP basic auth, or a `bearer` field to enable Bearer Token Authorization, or a `mTLS` object to enable Mutual TLS authentication                                     |
```

pkg/plugins/builtin/httputil/httputil.go
```diff
@@ -110,3 +113,42 @@ func defaultHTTPClientFactory(cfg HTTPClientConfig) HTTPClient {
...
+		caCertPool.AppendCertsFromPEM(caCert)
```
loopfz, line 151: it seems to fail silently ?

pkg/plugins/builtin/http/README.md
rbeuque74: thanks

pkg/plugins/builtin/httputil/httputil.go
rbeuque74: no, I will add error handling

rbeuque74:
Thanks for all the comments, all observations have been addressed (I guess).
Ready for final review.

pkg/plugins/builtin/http/http.go
```diff
@@ -110,6 +107,30 @@ func validConfig(config interface{}) error {
...
+	if cfg.Auth.Basic != nil && cfg.Auth.Bearer != nil {
+		return fmt.Errorf("basic auth and bearer auth are mutually exclusive")
+	}
+
+	if cfg.Auth.Basic != nil {
+		if cfg.Auth.Basic.User != "" && cfg.Auth.Basic.Password != "" {
+			// all good
+		} else {
+			return fmt.Errorf("missing either user or password for basic auth")
+		}
```
Alkorin, line 119: Why not: ?
```suggestion
		if cfg.Auth.Basic.User == "" || cfg.Auth.Basic.Password == "" {
			return fmt.Errorf("missing either user or password for basic auth")
		}
```

pkg/plugins/builtin/http/http.go
```diff
@@ -110,6 +107,30 @@ func validConfig(config interface{}) error {
...
+	if cfg.Auth.Bearer != nil && *cfg.Auth.Bearer == "" {
+		return fmt.Errorf("missing bearer token value")
+	}
+
+	if cfg.Auth.MutualTLS != nil {
+		if cfg.Auth.MutualTLS.ClientCert != "" && cfg.Auth.MutualTLS.ClientKey != "" {
+			// all good
+		} else {
+			return fmt.Errorf("missing either client_cert or client_key for mTLS")
+		}
```
Alkorin, line 131:
```suggestion
		if cfg.Auth.MutualTLS.ClientCert == "" || cfg.Auth.MutualTLS.ClientKey == "" {
			return fmt.Errorf("missing either client_cert or client_key for mTLS")
		}
```

loopfz (dismissed):

pkg/plugins/builtin/httputil/httputil.go
```diff
@@ -110,3 +113,58 @@ func defaultHTTPClientFactory(cfg HTTPClientConfig) HTTPClient {
...
+	}
+}
+
+// WithTLSRootCA should be called only once, with multiple PEM encoded certificates as input if needed.
+func WithTLSRootCA(caCert []byte) func(*http.Transport) error {
+	return func(t *http.Transport) error {
+		if t.TLSClientConfig == nil {
+			t.TLSClientConfig = &tls.Config{}
+		}
+		caCertPool, err := x509.SystemCertPool()
```
Alkorin, line 157: This code will not works on windows

rbeuque74: Yes, it will not work on Windows, that's why the error `if` statement just after handle the Windows case, and will have an empty pool to prevent an hard fail.

wI2L (approved):

PR #207, opened 2020-12-16
fix: sql migration 1.10.0: handle null last_start on resolution

rbeuque74:
Fix SQL migration

Unable to apply migration

When some resolution are created before never run (for example, due to
TO_AUTORUN_DELAYED, or when a manual resolving is needed), last_start
property is NULL, which is incompatible with the NOT NULL statement for
created property. This ALTER corrects the issue.

No

* **Other information**:
Mandatory for 1.10 release


rbeuque74:
Related to #192

amstuta (approved):

PR #210, opened 2020-12-20
templating: adding fieldFrom helper function to retrieve a field from a previous expression

### LINKED ISSUE #200: templating: field operator pipe support
**Describe the solution you'd like**
`field` templating function only looks inside the templating global context when called.
In case we want to use it in a pipe context, we should take as the first parameter the context on which to apply.

Example: (context)
```yaml
"step":
  "first":
    "output": 
       "result": 
           "my-payload": "{\"common-name\":\"utask.example.org\",\"id\":32}"
```

To retrieve the `common_name` key, we could use templating like this:
```
{{ field `step` `first` `output` `result` `my-payload` | fromJson | fieldFrom `common-name` }}
```

The `fieldFrom` operator would take the context from the first argument, instead of the global task/resolution context; hence enabling the pipe operator to work as expected.

rbeuque74:
`field` templating helper function allow to retrieve a key into a
map[string], even if the key is not a valid templating keyword.
`fieldFrom` provides the same feature, expect that it doesn't use the
values map from the task/resolution/steps, but use the previous
templating pipeline as a templating source.

Closes #200

Signed-off-by: Romain Beuque <556072+rbeuque74@users.noreply.github.com>


loopfz:
Patch the list of templating funcs in README.md maybe?

PR #214, opened 2020-12-29
pkg/notify: add notification strategies regarding template name and state

rbeuque74:
Feature

Notification strategies allows µTask administrator to control when the
notification should be sent to a given notification backend. It can
control if the notification should be sent regarding the template used
by a task, or regarding the current state of the task.

No

* **Other information**:


loopfz (approved):
lgtm

PR #216, opened 2021-01-05
implements a digest auth on the http builtin and allow generate uuid …

ybriffa:
…in templates through a native method

It introduces the digest authentication method for the HTTP builtin and the uuid method to generate uuids in templates

None of the features exists, if we want to use any of them we need to use a builtin script

New features are well implemented

No

* **Other information**:
Happy new year

### DIFF
--- a/pkg/plugins/builtin/http/http.go
+++ b/pkg/plugins/builtin/http/http.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ovh/utask/pkg/plugins/builtin/httputil"
 	"github.com/ovh/utask/pkg/plugins/taskplugin"
 	"github.com/ovh/utask/pkg/utils"
+	dac "github.com/ybriffa/go-http-digest-auth-client"
 )
 
 // the HTTP plugin performs an HTTP call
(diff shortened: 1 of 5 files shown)


pkg/plugins/builtin/http/http.go
```diff
@@ -57,9 +58,10 @@ type parameter struct {
 
 // auth represents HTTP authentication
 type auth struct {
-	Basic     *authBasic `json:"basic"`
-	Bearer    *string    `json:"bearer"`
-	MutualTLS *mTLS      `json:"mutual_tls"`
+	Basic     *authBasic  `json:"basic"`
+	Digest    *authDigest `json:"digest"`
```
rbeuque74, line 62: This new authentication would need to be documented inside the plugin README: https://github.com/ovh/utask/blob/master/pkg/plugins/builtin/http/README.md

rbeuque74 (approved):

rbeuque74:
LGTM

@loopfz  @wI2L  any thoughts?

Btw, travis build is failing because we hit the `docker pull` limitation. I will disable travis-ci builds as they will EOL pretty soon (closing their free service), and we will only keep CDS build (cf #218)

wI2L (approved):

PR #277, opened 2021-05-28
fix: race condition inside an expanded step regarding dependencies

rbeuque74:
Bug fix

Children steps Dependencies array was shared between all the child steps
and the parent step, every modification to one child steps was
propagated to the other child. Impact was detected on template steps
with `foreach_strategy` set to `sequence` as the child steps might be
launched all together, instead of waiting the previous one.

No

* **Other information**:

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/ovh/utask"
 	"github.com/ovh/utask/engine/step"
+	"github.com/ovh/utask/engine/step/condition"
 	"github.com/ovh/utask/engine/values"
 	"github.com/ovh/utask/models/resolution"
 	"github.com/ovh/utask/models/runnerinstance"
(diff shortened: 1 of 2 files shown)


engine/engine.go
```diff
@@ -738,6 +739,18 @@ func expandStep(s *step.Step, res *resolution.Resolution) {
 	// generate all children steps
 	for i, item := range items {
 		childStepName := fmt.Sprintf("%s-%d", s.Name, i)
+
+		// copy all slices from the parent step to prevent array pointer
```
loopfz, line 743: Can we do it only when we follow a `foreach sequence` ?
I don't know of any other situation where we alter the children, and for big expands it could generate a lot of useless garbage.
If you're worried about maintainability, we could have a helper func e.g. `step.ShouldDeepCopy()` ?

loopfz (changes requested):

engine/engine.go
rbeuque74: I'm not that concerned about the garbage that it could creates, I expect all arrays to be relatively small.
Should we proceed with the fix of the bug, and check for alternative optimization later?

wI2L: Agreed with @rbeuque74, I think we can left as-is for the time being, we're not deling with huge arrays.

wI2L (approved):

engine/engine.go
rbeuque74: We can't wait anymore, let's prod this

PR #301, opened 2021-11-30
fix: go-staticcheck: break loop were not positionned at the correct level

rbeuque74:
Bug fix

At utask interruption, we were not processing correctly the break out of the loop, causing utask not to exit instantly.

utask is now exiting when it should.

No

* **Other information**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:latest AS js-builder
+FROM node:16 AS js-builder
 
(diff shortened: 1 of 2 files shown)


ovh-cds:
CDS Report package#1182.0 ✘  
* Stage 1
  * Docker build ✘

ovh-cds:
CDS Report package#1182.1 ✘  
* Stage 1
  * Docker build ✘

PR #302, opened 2021-11-30
fix: failure to commit leads to inconsistencies between workflow and …

rbeuque74:
Bug Fix

When a µTask worker fails to commit its current state to the database,
it was sometimes not stopping the execution of its resolution. This
leads to inconsistencies, as the step was currently RUNNING, but
database state indicates TODO. In that way, if the worker crashes, then
the new worker will restart with wrong information, inducing workflow
issues.

No

* **Other information**:

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -264,12 +264,14 @@ func (e Engine) launchResolution(publicID string, async bool, sm *semaphore.Weig
 	}
 	debugLogger.Debugf("Engine: Resolve() %s RECAP BEFORE resolve: state: %s, steps: %s", publicID, res.State, strings.Join(recap, ", "))
 	e.wg.Add(1)
+
+	err = nil
 	if async {
 		go resolve(dbp, res, t, sm, e.wg, debugLogger)
 	} else {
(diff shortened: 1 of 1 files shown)


ovh-cds:
CDS Report test#1183.0 ✘  
* tests and lint
  * golangci-lint ✘  

engine/engine.go
```diff
@@ -699,7 +715,9 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res
 				if s.State != step.StateAfterrunError {
 					res.SetStepState(s.Name, step.StateRunning)
 					step.PreRun(s, res.Values, resolutionStateSetter(res, preRunModifiedSteps), executedSteps)
-					commit(dbp, res, nil)
+					if err := commit(dbp, res, nil); err != nil {
+						return 0, err
```
rbeuque74, line 719: Problem in my current implementation is that exiting the `runAvailableSteps` here, will close the main loop, while some step might still be running.

Not sure how we should handle this.

ovh-cds:
CDS Report package#1184.0 ✘  
* Stage 1
  * Docker build ✘

PR #303, opened 2021-12-21
feat: api: add a route to resolve a templating expression

rbeuque74:
feature

When debugging a failed task, we often need to resolve templated value
that we struggle to inspect. POST /resolution/:id/templating will
execute a templating expression given as input, and returns the resolved
expression as output. This route can only be used by admins.

No

* **Other information**:


ovh-cds:
CDS Report test#1193.0 ✘  
* tests and lint
  * golangci-lint ✘  

utask.go
```diff
@@ -386,3 +387,76 @@ func Config(store *configstore.Store) (*Cfg, error) {
 
 	return global, nil
 }
+
+// Init launches the task orchestration engine, providing it with a global context
```
wI2L, line 391: Wrong comment.

api/handler/resolution.go
```diff
@@ -901,3 +902,80 @@ func UpdateResolutionStepState(c *gin.Context, in *updateResolutionStepStateIn)
...
+	PublicID    string `path:"id" validate:"required"`
+	TemplateStr string `json:"template_str" validate:"required"`
+	StepName    string `json:"step_name"`
+}
+
+// ResolveTemplatingResolutionOut is the output of the HTTP route
+// for ResolveTemplatingResolution
+type ResolveTemplatingResolutionOut struct {
+	Result string  `json:"result"`
+	Error  *string `json:"error"`
```
wI2L, line 916: `omitempty` ?

api/handler/resolution.go
```diff
@@ -901,3 +902,80 @@ func UpdateResolutionStepState(c *gin.Context, in *updateResolutionStepStateIn)
 
 	return nil
 }
+
+type resolveTemplatingResolutionIn struct {
+	PublicID    string `path:"id" validate:"required"`
+	TemplateStr string `json:"template_str" validate:"required"`
```
wI2L, line 908: I suggest renaming `TemplateStr` to `Template`, as the type is already exposed.

rbeuque74: I think using `Template` might be confusing with the `tasktemplate`/`template_name` used a lot in µTask.
I suggest `TemplatingExpression`, what do you think ?

wI2L: Sounds good for `TemplatingExpression`, with field named `template_expr` maybe ?

rclsilver (approved):

richardlt (approved):

wI2L:
Note: this route should be restricted to admin users, to avoid leaking sensitive informations from resolutions.

PR #335, opened 2022-05-09
fix: mark resolution as waiting instead of deadlock

rclsilver:
bug fix

Currently, when we have a step which is a **subtask** in `WAITING` state, the resolution is **_dead locked_**.

With this new version, the resolution will be `WAITING` as expected.


rbeuque74 (approved):
Good catch

PR #343, opened 2022-07-23
fix(steps): provide functions args in function conditions values

rclsilver:
Bug fix

Currently, we don't have the function args values available in functions conditions (skip and check).

Now, the function args are available in functions conditions (skip and check).

No breaking change.

* **Other information**:


ovh-cds:
CDS Report test#1310.0 ✘  
* tests and lint
  * golangci-lint ✔ 
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1309.0 ✘  
* tests and lint
  * golangci-lint ✔ 
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1310.1 ✘  
* tests and lint
  * golangci-lint ✔ 
  * unit-test and integration-tests ✘

rbeuque74:
LGTM.
Need to check why the build is failing

rclsilver:
> LGTM. Need to check why the build is failing

I guess as I cloned the values, the changes made by the condition wasn't saved in the DB. I think we have to alter the original Values with the FunctionArgs.

What do you think?

rclsilver:
I added a test to verify my work

ovh-cds:
CDS Report test#1326.0 ✘  
* tests and lint
  * golangci-lint ✘  
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1327.0 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1327.1 ✘  
* tests and lint
  * golangci-lint ✘  
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1327.2 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1327.3 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1327.4 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1327.5 ✘  
* tests and lint
  * golangci-lint ✘  
  * unit-test and integration-tests ✘

engine/templates_tests/functionConditions.yaml
```diff
@@ -0,0 +1,29 @@
...
+  notExpected:
+    description: Say hello to foo but expect bar
+    action:
+      type: hello-with-conditions
+      configuration:
+        name: foo
+        expected: bar
+
+  skipped:
+    description: Say hello to foo but expect bar
```
rbeuque74, line 23:
```suggestion
    description: Say hello to foo but skipping step before it starts
```

ovh-cds:
CDS Report test#1333.0 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1332.0 ✘  
* tests and lint
  * golangci-lint ✘  

ovh-cds:
CDS Report test#1334.0 ✘  
* tests and lint
  * golangci-lint ✘  
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1335.0 ✘  
* tests and lint
  * golangci-lint ✘  
  * unit-test and integration-tests ✘

engine/functions_tests/nested2.yaml
```diff
@@ -22,6 +22,9 @@ conditions:
         - value: '{{.step.this.state}}'
           operator: EQ
           expected: 'NESTED2_STATE1'
+        - value: '{{ .function_args.name1 }}'
+          operator: EQ
+          expected: foobar
```
rbeuque74, line 27: I added this condition in the test : function `nested2` has correctly been called with `foobar`, but it's failing in the test.
This comes from the fact that the pull request is only using the function args of the "main" step (the one declared in the template): if we start to have nested functions, templating start to fails, because only the `function_args` on the "main" step is populated in the values.

I also found why this test was failing if we used `values.Clone`, follow the second comment

engine/step/step.go
```diff
@@ -521,6 +538,23 @@ func AfterRun(st *Step, values *values.Values, ss StateSetter) {
...
+	}
+
+	_, isFunction := runner.(*functions.Function)
+	if isFunction {
+		var functionInput map[string]interface{}
+		if err := utils.JSONnumberUnmarshal(bytes.NewBuffer(st.Action.Configuration), &functionInput); err != nil {
+			ss(st.Name, StateServerError, err.Error())
+			return
+		}
+		values.SetFunctionsArgs(functionInput)
```
rbeuque74, line 555: Here, we previously used `values.Clone` to prevent alter on the `values` object of the caller.
But, in this function, we have a second parameter, which is `ss` for `StateSetter`: https://github.com/ovh/utask/blob/master/engine/step/step.go#L521 
Called from here: https://github.com/ovh/utask/blob/master/engine/engine.go#L431
definition of this function is here: https://github.com/ovh/utask/blob/master/engine/engine.go#L1064-L1071

Everytime a condition is matching, we need to alter the `state` of some steps, using `ss` (such as here https://github.com/ovh/utask/blob/master/engine/step/step.go#L551 )
And the call of the closure in `resolutionStateSetter` is using the parent `values` object to alter the current state.
So when the second condition is evaluated, and require that the first condition altered the state to match, we are using the cloned value, which has not been altered by the closure in `resolutionStateSetter`, which mean that the condition state will not match, making the test failing.

I guess not cloning the values is fine, but the drawback of function_args not working in conditions if they are nested seems a huge drawback that we should address before merging this one

rbeuque74 (changes requested):
See other comments

PR #346, opened 2022-07-24
fix(step): stop the step execution if the pre-hook has failed

orandin:

Bug fix.

When a pre-hook fails, the step execution continues as if the pre-hook was executed correctly. In addition, the `stepChan` will receive the same step twice, because the pre-hook error handling sends the step into `stepChan` and the step execution also sends it.

To reproduce the case, you can use the example [1] (see below). The two first steps (`getTime` & `sayFoo`) will be executed, but the `sayFoo` will be fully executed and received twice by `stepChan`. Consequence, the `getTime` step is no longer tracked by the engine.

The action of the step will be executed if the pre-hook was executed correctly and the `stepChan` doesn't receive the same step twice.

No breaking change.

* **Other information**:

[1] Example of template to reproduce the case:
*Requirement: no application listening on `http://localhost`*

```yaml
name:             hello-world-now-prehook
description:      Test prehook

title_format:     Say hello in {{.input.language}}
result_format:
  echo_message: '{{.step.sayHello.output.message}}'
  echo_when:    '{{.step.sayHello.output.when}}'

allowed_resolver_usernames:   []
allow_all_resolver_usernames: true
auto_runnable: true
blocked:       false
hidden:        false

variables:
- name: english-message
  value: Hello World!
- name: spanish-message
  expression: |-
    // a short javascript snippet
    var h = 'Hola';
    var m = 'mundo';
    h + ' ' + m + '!';

inputs:
- name: language
  description: The language in which you wish to greet the world
  legal_values: [english, spanish]
  optional: true
  default: english

steps:
  getTime:
    description: Get UTC time
    action:
      type: http
      configuration:
        url: http://worldclockapi.com/api/json/utc/now
        method: GET
  sayFoo:
    pre_hook:
      type: http
      configuration:
        method: "GET"
        url: "http://localhost/"
    description: Echo a greeting in your language of choice
    action:
      type: echo
      configuration:
        output:
          message: "foo"
  sayHello:
    description: Echo a greeting in your language of choice
    dependencies:
     - getTime
     - sayFoo
    action:
      type: echo
      configuration:
        output:
          message: >-
            {{if (eq .input.language "english")}}{{eval "english-message"}}
            {{else if (eq .input.language "spanish")}}{{eval "spanish-message"}}{{end}}
          when: '{{.step.getTime.output.currentDateTime}}'
```

### DIFF
--- a/engine/step/step.go
+++ b/engine/step/step.go
@@ -367,6 +367,7 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, stepValues *values.Val
 		return
 	}
 
+	var prehookFailed bool
 	var preHookWg sync.WaitGroup
 	if prehook != nil {
 		preHookExecution, err := st.generateExecution(*prehook, baseConfig, stepValues, shutdownCtx)
@@ -383,6 +384,7 @@ func Run(st *Step, baseConfig map[string]json.RawMessage, stepValues *values.Val
 
 			st.execute(preHookExecution, func(output interface{}, metadata interface{}, tags map[string]string, err error) {
 				if err != nil {
+					prehookFailed = true
 					st.State = StateFatalError
 					st.Error = fmt.Sprintf("prehook: %s", err)
 					go noopStep(st, stepChan)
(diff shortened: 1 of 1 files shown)


rbeuque74 (approved):
Nice catch, thanks !

PR #366, opened 2022-10-10
fix(garbage): subtasks are kept until the parent task is in a final state

orandin:

Bug fix

The subtasks are deleted even if the parent task is not done.

**Side effect**: If the parent task hasn't been executed before the deletion of the subtask, the parent task is in an inconsistent state. It can't determine the subtask state and continue the execution.

The subtasks are kept until the parent task is in a final state. 

No

* **Other information**:

### Config

**utask-cfg**
```json
{
    "completed_task_expiration": "1m"
}
```

### Scripts

**scripts/sleep.sh**
```bash
#!/bin/bash

/bin/sleep 60
```

### Templates

<details><summary>Click here to show templates</summary>
<p>

```yaml
name:             garbage-parent
description:      Test garbage (parent)

title_format:     Test garbage

allowed_resolver_usernames:   []
allow_all_resolver_usernames: true
auto_runnable: true
blocked:       false
hidden:        false

steps:
  createSubtask:
    description: "Create subtask"
    action:
      type: subtask
      configuration:
        template: garbage-subtask
  Waiting:
    dependencies:
      - createSubtask
    description: "Wait one minute"
    action:
      type: script
      configuration:
        file_path: sleep.sh
```

```yaml
name:             garbage-subtask
description:      Test garbage (subtask)

title_format:     Test garbage (subtask)

allowed_resolver_usernames:   []
allow_all_resolver_usernames: true
auto_runnable: true
blocked:       false
hidden:        true

steps:
  Hello:
    description: "Display a message"
    action:
      type: echo
      configuration:
        output:
          foo: 'bar'
```

</p>
</details>

### DIFF
--- a/engine/collector_garbage.go
+++ b/engine/collector_garbage.go
@@ -6,9 +6,11 @@ import (
 	"time"
 
 	"github.com/loopfz/gadgeto/zesty"
+
 	"github.com/ovh/utask"
 	"github.com/ovh/utask/db/pgjuju"
 	"github.com/ovh/utask/models/task"
+	"github.com/ovh/utask/pkg/constants"
 	"github.com/ovh/utask/pkg/now"
 )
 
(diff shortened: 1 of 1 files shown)


rbeuque74 (changes requested):
If I understand correctly the modification on the SQL query, the subtask is deleted if it doesn't have the subtask tags, or if the parent task is in the DONE,CANCELLED,WONTFIX states.

In some weird case, if the parent task is getting deleted BEFORE the subtask, the subtask will never be deleted. That would be a case we need to support IMO

PR #371, opened 2022-10-17
Some fixes for make test-docker

orandin:

Bug fix

`make test-docker` fails because the timeout has expired. 
use `go get` to install goveralls
use postgres:9.4 for tests 

`make test-docker` is fixed (the timeout is increased)
use `go install` to install goveralls
use postgres:14 for tests 

No

* **Other information**:

Longest test:
```
ok      github.com/ovh/utask/engine     29.305s coverage: 78.1% of statements
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,11 +67,11 @@ test-dev:
 
 test-travis:
 	# moving to another location to go get some packages, otherwise it will include those packages as dependencies in go.mod
-	cd ${HOME} && go get golang.org/x/tools/cmd/cover github.com/mattn/goveralls
+	cd ${HOME} && go go install github.com/mattn/goveralls@latest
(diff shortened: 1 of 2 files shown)


rclsilver (approved):

Makefile
```diff
@@ -67,11 +67,11 @@ test-dev:
 
 test-travis:
 	# moving to another location to go get some packages, otherwise it will include those packages as dependencies in go.mod
-	cd ${HOME} && go get golang.org/x/tools/cmd/cover github.com/mattn/goveralls
+	cd ${HOME} && go go install github.com/mattn/goveralls@latest
```
rbeuque74, line 70: Go go (gadgeto) install

rbeuque74 (changes requested):

PR #373, opened 2022-10-18
chore: allow to override db schema and docker network in tests

rclsilver:
chore

not possible to use a docker network during tests in docker

allow to use an existing docker network during tests and allow to override the db schema file path

no breaking change

* **Other information**:


rbeuque74 (approved):
LGTM

PR #376, opened 2022-10-27
feat(ssh): add timeout support for SSH plugin

### LINKED ISSUE #209: ssh: add timeout configuration
**Describe the solution you'd like**
Allow `timeout` configuration on `ssh` builtin plugin to control the maximum time that can be spent waiting for the end of the execution of the ssh command.

rbeuque74:
Feature

Add `timeout` to SSH plugin

No

* **Other information**:
Closes #209

### DIFF
--- a/pkg/plugins/builtin/ssh/ssh.go
+++ b/pkg/plugins/builtin/ssh/ssh.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 
 	"github.com/ovh/utask/pkg/plugins/builtin/scriptutil"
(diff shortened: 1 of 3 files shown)


rclsilver (changes requested):
We can't set a valid duration as the timeout field is not a string.

Could you also update the template-schema.json please?

rclsilver (approved):

pkg/plugins/builtin/ssh/ssh.go
```diff
@@ -76,6 +79,15 @@ func configssh(i interface{}) error {
 		return fmt.Errorf("ssh too many hops (max %d)", MaxHops)
 	}
 
+	if cfg.Timeout != "" {
+		if cfg.Timeout[0] == '-' {
```
wI2L, line 83: This is superfluous, it can be checked after the duration is parsed in a single check.

rclsilver:
Missing json-schema update

rclsilver (changes requested):
need to add the new parameter in the json-schema

pkg/plugins/builtin/ssh/ssh.go
```diff
@@ -226,19 +244,55 @@ trap printResultJSON EXIT
...
+		}
+	}(ch)
+
+	if executionTimeout.Seconds() > 0 {
+		ctx, cancelFunc := context.WithTimeout(context.Background(), executionTimeout)
+		defer cancelFunc()
+		go func() {
+			select {
+			case <-funcContext.Done():
+			case <-ctx.Done():
```
Inozuma, line 269: You can probably avoid the double context by simply closing the session after the timeout, instead of having the session closed eventually with the `defer`.

Inozuma (approved):

PR #385, opened 2022-11-15
fix(sql): set empty string as default compression alg

rclsilver:
bug fix

resolutions created before the compression feature cannot be loaded because the `steps_compression_alg` column is filled with `NULL` as value.

`NULL` value are replaced by empty strings, the column is not nullable anymore.

no

* **Other information**:
-


ovh-cds:
CDS Report test#1491.0 ✘  
* tests and lint
  * golangci-lint ✔ 
  * unit-test and integration-tests ✘

ovh-cds:
CDS Report test#1492.0 ✘  
* tests and lint
  * golangci-lint ✔ 
  * unit-test and integration-tests ✘

rbeuque74 (approved):

PR #395, opened 2022-11-22
fix(resolution): restore hex.Decode/Encode when loading/storing ciphered data from database

rbeuque74:
Signed-off-by: Romain Beuque <556072+rbeuque74@users.noreply.github.com>

Bug fix

` Failed to load resolution from public id: chacha20poly1305: message authentication failed`

Previously, we were using symmecrypt.DecryptMarshal to load data from the database, and we switched to symmecrypt.Decrypt with bf23fbb10b62bb487ac4ea01b1e519f85480e58b.
This changed the behaviour because DecryptMarshal was also calling hex.Decode on the ciphered text before doing the Decrypt. We had to restore this behaviour to read old data from our database.
Same for EncryptMarshal/Encrypt/hex.Encode

No

* **Other information**:

### DIFF
--- a/models/resolution/resolution.go
+++ b/models/resolution/resolution.go
@@ -168,7 +168,7 @@ func Create(dbp zesty.DBProvider, t *task.Task, resolverInputs map[string]interf
 		return nil, err
 	}
 
-	dst := make([]byte, 0, hex.EncodedLen(len(encryptedSteps)))
+	dst := make([]byte, hex.EncodedLen(len(encryptedSteps)))
 	hex.Encode(dst, encryptedSteps)
 	r.EncryptedSteps = encryptedSteps
 
@@ -251,7 +251,7 @@ func load(dbp zesty.DBProvider, publicID string, locked bool, lockNoWait bool) (
 		return nil, err
 	}
 
-	dst := make([]byte, 0, hex.DecodedLen(len(r.EncryptedSteps)))
+	dst := make([]byte, hex.DecodedLen(len(r.EncryptedSteps)))
 
 	// if we can't hex Decode, we might be in the case of a Resolution row in database that was
 	// created between the v1.21.1 and v1.21.3 that was bugged, and failed to hex Encode/Decode the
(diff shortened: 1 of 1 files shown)


models/resolution/resolution.go
```diff
@@ -166,6 +167,9 @@ func Create(dbp zesty.DBProvider, t *task.Task, resolverInputs map[string]interf
 	if err != nil {
 		return nil, err
 	}
+
+	dst := make([]byte, 0, hex.EncodedLen(len(encryptedSteps)))
```
Inozuma, line 171:
```suggestion
	dst := make([]byte, hex.EncodedLen(len(encryptedSteps)))
```

From the official documentation, it uses the size of the buffer: https://pkg.go.dev/encoding/hex#Encode

models/resolution/resolution.go
```diff
@@ -377,6 +391,8 @@ func (r *Resolution) Update(dbp zesty.DBProvider) (err error) {
 		return err
 	}
 
+	dst := make([]byte, 0, hex.EncodedLen(len(compressedSteps)))
```
Inozuma, line 394:
```suggestion
	dst := make([]byte, hex.EncodedLen(len(compressedSteps)))
```

models/resolution/resolution.go
```diff
@@ -247,7 +251,17 @@ func load(dbp zesty.DBProvider, publicID string, locked bool, lockNoWait bool) (
 		return nil, err
 	}
 
-	compressedSteps, err := models.EncryptionKey.Decrypt(r.EncryptedSteps, []byte(r.PublicID))
+	dst := make([]byte, 0, hex.DecodedLen(len(r.EncryptedSteps)))
```
Inozuma, line 254:
```suggestion
	dst := make([]byte, hex.DecodedLen(len(r.EncryptedSteps)))
```

models/resolution/resolution.go
```diff
@@ -247,7 +251,17 @@ func load(dbp zesty.DBProvider, publicID string, locked bool, lockNoWait bool) (
...
+
+	// if we can't hex Decode, we might be in the case of a Resolution row in database that was
+	// created between the v1.21.1 and v1.21.3 that was bugged, and failed to hex Encode/Decode the
+	// ciphered data. We need to keep backward compatibility for those, but this should not happen
+	// often.
+	// See https://github.com/ovh/utask/commit/bf23fbb10b62bb487ac4ea01b1e519f85480e58b and migration
+	// from symmecrypt.Key.DecryptMarshal to symmecrypt.Key.Decrypt
+	_, _ = hex.Decode(dst, r.EncryptedSteps)
```
Inozuma, line 262: Should probably check returned `error` and set `dst` to `r.EncryptedSteps` if there was an error.

Inozuma (changes requested):

models/resolution/resolution.go
wI2L: Note: the functions uses the byte slice indexes, and not append, so the slice must be of the proper length, enough capacity with a 0 length won't work.

Inozuma (approved):

wI2L (approved):

PR #485, opened 2023-10-12
fix: capture step correctly into loop

### LINKED ISSUE #480: Captured loop variable inside a goroutine
I'm sure it qualifies as bug so I opened a blank issue, but I noticed a weird usage of a loop variable captured in a goroutine at:  https://github.com/ovh/utask/blob/101b9974e03a6639f7cd6f5dc60d7d27713caa41/engine/engine.go#L706

When iterating over available steps, if a `foreach` step is followed by other steps (at least one), the last step will actually be pushed into the channel in place of the `foreach` step.

The `foreach` step should be run at some point, but as long as there is another step after it, it won't be run at all.

rbeuque74:
Bug fix

For loop do not capture anymore the variable

No

* **Other information**:
Resolves #480

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -703,7 +703,9 @@ func runAvailableSteps(dbp zesty.DBProvider, modifiedSteps map[string]bool, res 
 				// rebuild step dependency tree to include generated loop steps
 				res.BuildStepTree()
 				commit(dbp, res, nil)
-				go func() { stepChan <- s }()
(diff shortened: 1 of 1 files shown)


ovh-cds:
CDS Report test#1772.0 ✘  
* tests and lint
  * golangci-lint ✘  

rclsilver (approved):

PR #495, opened 2023-11-30
feat(plugin): add new task-batching plugin

DrRebus:
A new feature: a task batching plugin.

The only way of batching tasks so far is to use a `subtask` plugin with the `foreach` feature, which dynamically adds a new step per `subtask`. This approach isn't efficient DB-wise, because everytime a step is updated, the whole step tree (a JSON formatted string) gets updated, triggering an `UPDATE` in the database. These `UPDATE`s create new bigger-and-bigger rows and delete the previous ones, bloating the DB and requiring a stronger `VACUUM`.

The task-batching plugin spawns child tasks as independent tasks that wake the parent once completed. Since the parent task no longer embed its children, no hard link exists between them, but they're linked via:
- the `ParentTaskID` tag, the public ID of the parent task, allowing the child to wake its parent.
- the `BatchID` property, the public ID of the batch the task belongs two.

Nope.

* **Other information**:
This plugin doesn't yet aggregate results of the children tasks. It'll come in a subsequent version.
Contrary to what's been discussed, when using sub batches (`sub_batch_size` > 0), children tasks don't always wake their parent when they're done. Instead, the last task in the sub batch will wake the parent. That way, we limit even more the amount of UPDATEs made. The downside is that we have to wait for the whole sub batch to be done before spawning more tasks in the batch.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ovh/utask/pkg/jsonschema"
 	"github.com/ovh/utask/pkg/metadata"
 	"github.com/ovh/utask/pkg/now"
+	pluginbatch "github.com/ovh/utask/pkg/plugins/builtin/batch"
 	"github.com/ovh/utask/pkg/taskutils"
 	"github.com/ovh/utask/pkg/utils"
 )
@@ -524,7 +525,9 @@ forLoop:
 			if mapStatus[status] {
 				if status == resolution.StateWaiting && recheckWaiting {
 					for name, s := range res.Steps {
-						if s.State == step.StateWaiting {
+						// Steps using the batch plugin shouldn't be run again when WAITING. Running them second time
+						// may lead to a race condition when the last task of a sub-batch tries to resume its parent
+						if s.State == step.StateWaiting && s.Action.Type != pluginbatch.Plugin.PluginName() {
 							delete(executedSteps, name)
 						}
 					}
--- a/pkg/batch/batch.go
+++ b/pkg/batch/batch.go
@@ -1,0 +1,77 @@
+package batch
+
+import (
+	"context"
+
+	"github.com/juju/errors"
+	"github.com/loopfz/gadgeto/zesty"
+
+	"github.com/ovh/utask/models/task"
+	"github.com/ovh/utask/models/tasktemplate"
+	"github.com/ovh/utask/pkg/taskutils"
+)
+
+// TaskArgs holds arguments needed to create tasks in a batch
+type TaskArgs struct {
+	TemplateName     string                   // Mandatory
+	Inputs           []map[string]interface{} // Mandatory
+	CommonInput      map[string]interface{}   // Optional
+	Comment          string                   // Optional
+	WatcherUsernames []string                 // Optional
+	WatcherGroups    []string                 // Optional
+	Tags             map[string]string        // Optional
+}
+
+// Populate creates and adds new tasks to a given batch.
+// All tasks share a common batchID which can be used as a listing filter.
+// The [constants.SubtaskTagParentTaskID] tag can be set in the Tags to link the newly created tasks to another
+// existing task, making it the parent of the batch. A parent task is resumed everytime a child task finishes.
+func Populate(ctx context.Context, batch *task.Batch, dbp zesty.DBProvider, args TaskArgs) ([]string, error) {
+	tt, err := tasktemplate.LoadFromName(dbp, args.TemplateName)
+	if err != nil {
+		return nil, err
+	}
+
+	taskIDs := make([]string, 0, len(args.Inputs))
+	for _, inp := range args.Inputs {
+		input, err := mergeMaps(args.CommonInput, inp)
+		if err != nil {
+			return nil, err
+		}
+
+		t, err := taskutils.CreateTask(
+			ctx,
+			dbp,
+			tt,
+			args.WatcherUsernames,
+			args.WatcherGroups,
+			[]string{},
+			[]string{},
+			input,
+			batch,
+			args.Comment,
+			nil,
+			args.Tags,
+		)
+		if err != nil {
+			return nil, err
+		}
+		taskIDs = append(taskIDs, t.PublicID)
+	}
+	return taskIDs, nil
+}
+
+func mergeMaps(common, particular map[string]interface{}) (map[string]interface{}, error) {
+	merged := make(map[string]interface{}, len(common)+len(particular))
+	for key, value := range particular {
+		merged[key] = value
+	}
+
+	for key, value := range common {
+		if _, ok := merged[key]; ok {
+			return nil, errors.NewBadRequest(nil, "Conflicting keys in input maps")
+		}
+		merged[key] = value
+	}
+	return merged, nil
+}
--- a/pkg/batchutils/batchutils.go
+++ b/pkg/batchutils/batchutils.go
@@ -1,0 +1,28 @@
+package batchutils
+
+import (
+	"github.com/Masterminds/squirrel"
+	"github.com/loopfz/gadgeto/zesty"
+
+	"github.com/ovh/utask/db/sqlgenerator"
+	"github.com/ovh/utask/models/task"
+)
+
+// FinalStates hold the states in which a task won't ever be run again
+var FinalStates = []string{task.StateDone, task.StateCancelled, task.StateWontfix}
+
+// RunningTasks returns the amount of running tasks sharing the same given batchId.
+func RunningTasks(dbp zesty.DBProvider, batchID int64) (int64, error) {
+	query, params, err := sqlgenerator.PGsql.
+		Select("count (*)").
+		From("task t").
+		Join("batch b on b.id = t.id_batch").
+		Where(squirrel.Eq{"b.id": batchID}).
+		Where(squirrel.NotEq{"t.state": FinalStates}).
+		ToSql()
+	if err != nil {
+		return -1, err
+	}
+
+	return dbp.DB().SelectInt(query, params...)
+}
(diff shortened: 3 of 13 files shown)


engine/engine.go
```diff
@@ -524,7 +525,8 @@ forLoop:
 			if mapStatus[status] {
 				if status == resolution.StateWaiting && recheckWaiting {
 					for name, s := range res.Steps {
-						if s.State == step.StateWaiting {
+						// Steps using the batch plugin shouldn't be run again when WAITING
+						if s.State == step.StateWaiting && s.Action.Type != pluginbatch.Plugin.PluginName() {
```
DrRebus, line 529: I'm not sure why steps are run once again when they reach the WAITING state, but it creates a race condition with the batch plugin. If the last task of the a sub-batch tries to resume its parent, but said parent is being run a second time by the engine, preventing the child task from resuming it.

floriancazals (approved):

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,362 @@
...
+
+	taskIDs, err := batch.Populate(ctx, b, dbp, args)
+	if err != nil {
+		return []string{}, err
+	}
+
+	return taskIDs, nil
+}
+
+// runBatch runs batch, spawning new tasks if needed and checking whether they're all done.
```
wI2L, line 261:
```suggestion
// runBatch runs a batch, spawning new tasks if needed and checking whether they're all done.
```

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return metadata, err
+	}
+	metadata.RemainingTasks = running
+	return metadata, nil
+}
+
+// increaseRunMax increases the maximum amount of runs of the resolution matching the given parentTaskID by the run
+// count of the given batchStepName.
+// Since child tasks wake their parent up when they're done, the resolution's RunCount gets incremented everytime. We
```
wI2L, line 305:
```suggestion
// Since child tasks wake up their paren when they're done, the resolution's RunCount gets incremented everytime. We
```

pkg/taskutils/taskutils.go
```diff
@@ -88,6 +89,24 @@ func ShouldResumeParentTask(dbp zesty.DBProvider, t *task.Task) (*task.Task, err
 		return nil, nil
 	}
 
+	if t.BatchID != nil {
+		// The task belongs to a batch. If all sibling tasks are done, the parent can be awaken.
+
+		// Note on race conditions:
+		// When two sibling tasks complete, they either complete at the very same time or with a delay. In the former
+		// case, two attempts to resume the parent may be triggered, but a DB lock already prevents the parent
+		// from being run twice at the same time. In the later case, no race condition exists since finished before
```
wI2L, line 98:
```suggestion
		// from being run twice at the same time. In the later case, no race condition exists since one finished before
```

pkg/batchutils/batchutils.go
```diff
@@ -0,0 +1,28 @@
...
+
+// States in which the task won't ever be run again
+var FinalStates = []string{task.StateDone, task.StateCancelled, task.StateWontfix}
+
+// RunningTasks returns the amount of running tasks sharing the same given batchId.
+func RunningTasks(dbp zesty.DBProvider, batchId int64) (int64, error) {
+	query, params, err := sqlgenerator.PGsql.
+		Select("count (*)").
+		From("task t").
+		Join("batch b on b.id = t.id_batch").
```
wI2L, line 19: `id_batch` should be `batch_id` to be coherent with the rest of the code, and it reads better.

pkg/batch/batch.go
```diff
@@ -0,0 +1,77 @@
...
+		)
+		if err != nil {
+			return []string{}, err
+		}
+		taskIDs = append(taskIDs, t.PublicID)
+	}
+	return taskIDs, nil
+}
+
+func conjMap(common, particular map[string]interface{}) (map[string]interface{}, error) {
```
wI2L, line 64: The name of this function does not clearly reflect its usage.
It does a merge/union of two map while checking that common keys are not overriden => btw, I'm skeptical of the check line 71/72. It seems legitimate to want to define a "default/common" value for all tasks, while still overriding it for some tasks, no?

wI2L (changes requested):

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,362 @@
...
+	var stepError error
+
+	conf := config.(*BatchConfig)
+	batchCtx := ictx.(*BatchContext)
+	if err := parseInputs(conf, batchCtx); err != nil {
+		return nil, batchCtx.RawMetadata.Format(), err
+	}
+
+	if conf.Tags == nil {
+		conf.Tags = map[string]string{}
```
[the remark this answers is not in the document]
DrRebus: Is it a question of style or is there a specific reason?

pkg/plugins/builtin/batch/batch.go
DrRebus: I'm pretty sure "waking [someone / something] up" is correct. :upside_down_face:
Though, both seem fine: https://dictionary.cambridge.org/fr/dictionnaire/anglais/wake-up

pkg/batchutils/batchutils.go
DrRebus: I couldn't agree more, but `id_batch` is the column's name in the database, I have no power over it sadly. :smiling_face_with_tear:
https://github.com/ovh/utask/blob/master/sql/schema.sql#L44

pkg/batch/batch.go
DrRebus: This function used to sit elsewhere (https://github.com/ovh/utask/blob/master/api/handler/batch.go#L80), I simply moved it and kept the exact same behaviour. Changing its name (something like `mergeMaps` seems adequate) won't have any side effect, but I'm not sure changing its behaviour is 100% safe.
That being said, allowing each task to override the default value is a good idea and seems like a valid use case.

engine/engine.go
orandin: Add this explanation in comment to understand this quickfix/hack

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,362 @@
...
+// increaseRunMax increases the maximum amount of runs of the resolution matching the given parentTaskID by the run
+// count of the given batchStepName.
+// Since child tasks wake their parent up when they're done, the resolution's RunCount gets incremented everytime. We
+// compensate this by increasing the RunMax property once the batch is done.
+func increaseRunMax(dbp zesty.DBProvider, parentTaskID string, batchStepName string) error {
+	t, err := task.LoadFromPublicID(dbp, parentTaskID)
+	if err != nil {
+		return err
+	}
+	res, err := resolution.LoadLockedFromPublicID(dbp, *t.Resolution)
```
orandin, line 312: **Style**: Add new line

```suggestion

	res, err := resolution.LoadLockedFromPublicID(dbp, *t.Resolution)
```

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return err
+	}
+	step, ok := res.Steps[batchStepName]
```
orandin, line 316: **Style**: add new line


```suggestion

	step, ok := res.Steps[batchStepName]
```

orandin (changes requested):

orandin (approved):

pkg/plugins/builtin/batch/README.md
```diff
@@ -13,10 +13,10 @@ The output of child tasks is not made available in this plugin's output. This fe
 | `template_name`      | the name of a task template, as accepted through µTask's API                                                      |
 | `inputs`             | a list of mapped key/value, as accepted on µTask's API. Each element represents the input of an individual task   |
 | `json_inputs`        | same as `inputs`, but as a JSON string. If specified, it overrides `inputs`                                       |
-| `common_input`       | a map of named values, as accepted on µTask's API, given to all task in the batch by combining it with each input |
-| `json_common_input`  | same a `common_input` but as a JSON string. If specified, it overrides `common_input`                             |
+| `common_inputs`       | a map of named values, as accepted on µTask's API, given to all task in the batch by combining it with each input |
+| `common_json_inputs`  | same a `common_inputs` but as a JSON string. If specified, it overrides `common_inputs`                             |
```
wI2L, line 17: s/same a/same as

pkg/plugins/builtin/batch/batch.go
```diff
@@ -0,0 +1,364 @@
...
+// count of the given batchStepName.
+// Since child tasks wake their parent up when they're done, the resolution's RunCount gets incremented everytime. We
+// compensate this by increasing the RunMax property once the batch is done.
+func increaseRunMax(dbp zesty.DBProvider, parentTaskID string, batchStepName string) error {
+	t, err := task.LoadFromPublicID(dbp, parentTaskID)
+	if err != nil {
+		return err
+	}
+
+	res, err := resolution.LoadLockedFromPublicID(dbp, *t.Resolution)
```
wI2L, line 313: Need to check for `t.Resolution == nil` before using it.

wI2L (changes requested):

wI2L (approved):

ovh-cds:
CDS Report test#1912.0 ✘  
* 
  * golangci-lint ✘  

ovh-cds:
CDS Report package#1913.0 ✘  
* Stage 1
  * Docker build ✘

wI2L (approved):

Issue #499, opened 2024-01-10
Steps' conditions are evaluated even when in a non-final state.


DrRebus:
**Describe the bug**
Steps' conditions are evaluated even though the step didn't reach a state that won't change anymore (i.e.: a "final" state), like DONE. Since these conditions can change the step's state, this state may change even when the step isn't fully executed. Whether or no to evaluate such conditions is defined at: https://github.com/ovh/utask/blob/master/engine/step/step.go#L534 and looks like:
```go
func AfterRun(st *Step, values *values.Values, ss StateSetter) {
	if st.skipped || st.State == StateServerError || st.State == StateFatalError || st.ForEach != "" {
		return
	}
	// Evaluate step's conditions
}
```
When a step uses the `subtask` plugin for instance, the step falls in the WAITING state until the spawned task finishes, but its conditions are still evaluated when it reached this state. Since the spawned task isn't DONE, its result may be wrong (e.g.: unevaluated templated results), so a step condition checking the subtask's result will use wrong values.

**To Reproduce**
1. Create a task template with a `result_format` containing at least one key, like:
```yaml
name: subtask-test-condition
title_format: 'Subtask test condition'
description: 'Template to be called as a subtask'

auto_runnable: true

result_format:
  some_key: '{{ .step.SomeStep.output }}'

steps:
  SomeStep:
    description: Some step
    action:
      type: echo
      configuration:
        output: "true"
```
2. Create another task template that will call the previous template in a subtask and add a check condition on the subtask's result, like so:
```yaml
name: test-step-condition
title_format: 'Test step condition'
description: 'Test step condition'

auto_runnable: true

steps:
  CallingSubtask:
    description: Calling the subtask
    custom_states:
      - MY_CUSTOM_STATE
    action:
      type: subtask
      configuration:
        template: subtask-test-condition
    conditions:
      - type: check
        if:
          - value: '{{ .step.this.output.result.some_key }}'
            operator: NE
            expected: "true"
        then:
          this: MY_CUSTOM_STATE
```
3. Check the step's state. It shouldn't be MY_CUSTOM_STATE since the subtask returns "true", yet it is.

**Expected behavior**
Steps' conditions should only be evaluated when a step reached a final state, like DONE.

DrRebus:
A way to work around this issue is to also check the subtask's state before checking its result, like:
```yaml
    conditions:
      - type: check
        if:
          - value: '{{ .step.this.output.state }}'
            operator: EQ
[...]

PR #500, opened 2024-01-11
fix(step): Prevented step's conditions from being evaluated when the step isn't in a final state

DrRebus:
Bug fix.

Steps' conditions are evaluated even though the step isn't in a final state. See #499.

Steps' conditions evaluation is also prevented when the step is in the WAITING, RUNNING or TO_RETRY state.

No.

### DIFF
--- a/engine/step/step.go
+++ b/engine/step/step.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -65,6 +66,7 @@ var (
 	stepConditionValidStates = []string{StateDone, StatePrune, StateToRetry, StateRetryNow, StateFatalError, StateClientError}
 	runnableStates           = []string{StateTODO, StateServerError, StateClientError, StateFatalError, StateCrashed, StateToRetry, StateRetryNow, StateAfterrunError, StateExpanded, StateWaiting} // everything but RUNNING, DONE, PRUNE
 	retriableStates          = []string{StateServerError, StateToRetry, StateAfterrunError}
+	validAfterRunStates      = []string{StateDone, StateClientError, StateAfterrunError}
 )
 
 // Step describes one unit of work within a task, and its dependency to other steps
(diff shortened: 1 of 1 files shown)


engine/step/step.go
```diff
@@ -531,7 +531,11 @@ func PreRun(st *Step, values *values.Values, ss StateSetter, executedSteps map[s
 // AfterRun evaluates a step's "check" conditions after the Step's action has been performed
 // and impacts the entire task's execution flow through the provided StateSetter
 func AfterRun(st *Step, values *values.Values, ss StateSetter) {
-	if st.skipped || st.State == StateServerError || st.State == StateFatalError || st.ForEach != "" {
+	switch st.State {
```
wI2L, line 534: I believe an inverse condition that allows to continue the execution when the step is in a "final" state would be clearer, and probably would avoid an issue if a new intermediate/error state is introduced in the future,

DrRebus: Right. I initially refrained from inverting this condition, specifying only "final" states, because custom states are considered "final" and we obviously can't list all custom states. Though, custom states are set by conditions, so it's a non-issue.
Still, if new "final" states are ever added, we would hit the same problem, wouldn't we?

orandin: How about checking if the step state is not present in the slice `stepConditionValidStates`? This would resolve issues if new state is added in the future.

```suggestion
	if st.skipped || st.ForEach != "" || !slices.Contains(stepConditionValidStates, st.State) {
```

orandin: How about checking if the step state is not present in slice `runnableStates`? This would resolve issues if new state is added in the future.

```suggestion
	if st.skipped || st.ForEach != "" || !slices.Contains(runnableStates, st.State) {
```

DrRebus: We can't simply use `runnableStates` because some error states (like `CLIENT_ERROR`) are runnable but shouldn't prevent conditions from being executed. For instance, a step performing an HTTP call that returns a 4XX status code sets the step in the `CLIENT_ERROR` state, state that we would override in a condition because we expected to receive a 4XX.

wI2L (approved):

PR #512, opened 2024-03-01
fix: forward error message when function is not existing

rbeuque74:
Bug fix

Error message obsfuscated

Error message visible

NO

* **Other information**:
Do it

### DIFF
--- a/engine/step/stepcondition.go
+++ b/engine/step/stepcondition.go
@@ -1,6 +1,8 @@
 package step
 
 import (
+	"fmt"
+
 	"github.com/juju/errors"
 	"github.com/ovh/utask/engine/step/condition"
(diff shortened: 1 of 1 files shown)


engine/step/stepcondition.go
```diff
@@ -42,7 +44,7 @@ func ValidCondition(sc *condition.Condition, stepName string, steps map[string]*
 
 		customStates, err := impactedStep.GetCustomStates()
 		if err != nil {
-			return errors.New("Step custom states are invalid")
+			return fmt.Errorf("Step custom states are invalid: %s", err)
```
wI2L, line 47:
```suggestion
			return fmt.Errorf("invalid custom stats found for step %s: %s", stepName, err)
```

rclsilver (approved):

PR #543, opened 2024-12-03
fix(plugin): Fixed batch plugin stuck when underlying batch is collected

DrRebus:
Bug fix and an improvement for the batch plugin

Steps using the batch plugin sometimes get stuck in the "WAITING" state because their underlying batch (in the database) was garbage-collected before they were able to populate it again.

The batch plugin no longer stay stuck in the "WAITING" state; it re-creates the batch with the same public ID and populates it.

Yes, the `sub_batch_size` parameter of the batch plugin now expects a string instead of a number to support both hardcoded values (e.g.: "5") and templated ones (e.g.: "{{ .config.myconfig.batch_size }}").

* **Other information**:

### DIFF
--- a/pkg/plugins/builtin/batch/README.md
+++ b/pkg/plugins/builtin/batch/README.md
@@ -1,9 +1,10 @@
 # `batch` Plugin
 
-This plugin creates a batch of tasks based on the same template and waits for it to complete. It acts like the `subtask` combined with a `foreach`, but doesn't modify the resolution by adding new steps dynamically. As it makes less calls to the underlying database, this plugin is suited for large batches of tasks, where the `subtask` / `foreach` combination would usually struggle, escpecially by bloating the database.
-Tasks belonging to the same batch share a common `BatchID` as well as tag holding their parent's ID.
+This plugin creates a batch of tasks based on the same template and waits for it to complete. It acts like the `subtask` combined with a `foreach`, but doesn't modify the resolution by adding new steps dynamically. As it makes less calls to the underlying database, this plugin is suited for large batches of tasks, where the `subtask` / `foreach` combination would usually struggle, especially by bloating the database.
+Tasks belonging to the same batch share a common `BatchID` as well as a tag holding their parent's ID.
 
 ##### Remarks:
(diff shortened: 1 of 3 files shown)


pkg/plugins/builtin/batch/README.md
```diff
@@ -1,9 +1,10 @@
...
-Tasks belonging to the same batch share a common `BatchID` as well as tag holding their parent's ID.
+This plugin creates a batch of tasks based on the same template and waits for it to complete. It acts like the `subtask` combined with a `foreach`, but doesn't modify the resolution by adding new steps dynamically. As it makes less calls to the underlying database, this plugin is suited for large batches of tasks, where the `subtask` / `foreach` combination would usually struggle, especially by bloating the database.
+Tasks belonging to the same batch share a common `BatchID` as well as a tag holding their parent's ID.
 
 ##### Remarks:
+Like the subtask plugin, it's unadvised to have a step based on the batch plugin running alongside other steps in a template. If these other steps take time to return a result, the batch plug may miss the wake up call from its children tasks.
```
wI2L, line 7:
```suggestion
Like the subtask plugin, it's unadvised to have a step based on the batch plugin running alongside other steps in a template. If these other steps takes time to return a result, the batch plugin may miss the wake up call from its children tasks.
```

pkg/plugins/builtin/batch/batch.go
```diff
@@ -46,8 +48,9 @@ type BatchConfig struct {
 	Tags              map[string]string        `json:"tags"`
 	ResolverUsernames string                   `json:"resolver_usernames"`
 	ResolverGroups    string                   `json:"resolver_groups"`
-	// How many tasks will run concurrently. 0 for infinity (default)
-	SubBatchSize int `json:"sub_batch_size"`
+	// How many tasks will run concurrently. 0 for infinity (default). It's supplied as a string to support templating
+	SubBatchSizeStr string `json:"sub_batch_size"`
+	SubBatchSize    int64
```
wI2L, line 53: unexport or add `json:"-"`

wI2L (approved):

PR #561, opened 2025-04-09
fix: Fixed tasks using Batch or Subtask plugin getting stuck in the WAITING state

DrRebus:
Bug Fix

Tasks using the Batch or Subtask plugin sometimes get stuck in the WAITING state because they miss the child task wake-up event.
When a child task finishes, it attempts to wake its parent up by running it. If the parent is already running at this time, this event is ignored and the child task terminates. However, the parent task may have run the step related to the child task before it finished and may think the child task is still running.

Instead of trying to run the parent, child tasks attempt to change the parent's `next_retry` date so they get rerun as soon as possible.

Nope.

* **Other information**:


orandin (approved):

wI2L (dismissed):

wI2L:
@DrRebus Please fix CI errors.

PR #631, opened 2026-04-29
fix: Fixed resolutions reaching BLOCKED_DEADLOCK when a Step's state is modified externally

DrRebus:
Bug fix.

Resolutions can end up in the `BLOCKED_DEADLOCK` state when one of their Steps is updated through the API because no pruning is performed.
Let's consider a simple task with 3 steps : `Parent`, `Child1` and `Child2`. `Child1` has the dependency `Parent:DONE`, `Child2` has `Parent:CUSTOM_STATE` (see attached picture). If `Parent` has an issue and can't reach either the `DONE` or `CUSTOM_STATE` and we were to update its state manually to the `DONE` state for example, `Child1` could be executed and `Child2` should reach the `PRUNE` state.  However, since no pruning is performed at this stage, `Child2` stays in `TODO` forever and the resolution becomes `BLOCKED_DEADLOCK`

Proactive pruning is performed whenever a resolution is run.

Not that I am aware of.

* **Other information**:
image

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -359,6 +359,15 @@ func initialize(dbp zesty.DBProvider, publicID string, debugLogger *logrus.Entry
 		return nil, nil, nil
 	}
 
+	{
+		// Making sure to prune steps whose dependencies were updated outside the engine (API, manually in DB, etc)
+		stepsToCheck := map[string]bool{}
+		for stepName := range res.Steps {
+			stepsToCheck[stepName] = true
+		}
+		pruneSteps(res, stepsToCheck)
+	}
+
 	if err := dbp.Commit(); err != nil {
 		dbp.Rollback()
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -359,6 +359,15 @@ func initialize(dbp zesty.DBProvider, publicID string, debugLogger *logrus.Entry
 		return nil, nil, nil
 	}
 
+	{
+		// Making sure to prune steps whose dependencies were updated outside the engine (API, manually in DB, etc)
```
DrRebus, line 363: We can also perform proactive pruning in the API's handlers whenever a resolution or step is updated, but doing it here also covers updates made in the database directly.

rclsilver (approved):

